PR #150, opened 2015-03-09
adding 2d projection to mat3

andrevenancio:
While developing a small framework to work with Quads on 2D environment, and for speed purposes, I found it useful to add a project2d style method. It receives the canvas width and height, and inverts on the Y axis so the 0 coordinate is on top.


andrevenancio:
actually canceling it and moving it to the dev branch and adding tests.. will re issue a PR soon

PR #154, opened 2015-04-09
Update ruby dependencies for Ruby >= 2.2.

jenanwise:
Previous version did not specify "rspec" in the Gemfile, which caused `rake` (as specified in TESTING.md) to not work after a `bundle install`.

Previous version pinned to json 1.7.6, which is incompatible with Ruby >= 2.2. New version is autogenerated as of 2015-04-09 and works with the supplied rake tasks. 

Previous version used `:gemcutter` for the gems source which uses http. This was switched to 'https://rubygems.org' as recommended by bundler docs.

Note that I don't really know Ruby, and the checked-in `Gemfile.lock` seems a bit crazy to me (especially without pinned versions in Gemfile), so I don't know the protocol on when it should be updated or not. But, this makes the tasks work again on an up-to-date OSX with Ruby 2.2.1 as of the time of writing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source :gemcutter
+source 'https://rubygems.org'
 
 # core ruby
 gem 'rake'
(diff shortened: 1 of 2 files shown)


jenanwise:
Out of date.

PR #155, opened 2015-04-09
Make versions consistent across source files.

jenanwise:
Previously, 2.1.0, 2.2.0, and 2.3.0 were all specified in different places.

This commit does NOT update the generated `dist/` files.


jenanwise:
Out of date.

PR #157, opened 2015-04-22
Implement vec4.random as a vector uniformly distributed on a 4-sphere surface

joewreschnig:
`vec4.random` has had a TODO to improve it for some time; it generated uniformly distributed elements and then normalized and scaled the result. This is projecting a hypercube onto a hypersphere, and like projecting a cube onto a sphere, the distribution has dense "corners."

This changes it to follow [Marsaglia (1972)](http://projecteuclid.org/euclid.aoms/1177692644) and generate the points uniformly on the hypersphere's surface. It is slightly less efficient because of the rejected samples, but still contains only one sqrt call.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -25,4 +25,4 @@ If Ruby is not installed, you must test with `jasmine-node` directly:
 
     NODE_PATH=$NODE_PATH:.                       \
       node_modules/jasmine-node/bin/jasmine-node \
-      spec/helpers/node_helper.js
+      spec/helpers/node-helper.js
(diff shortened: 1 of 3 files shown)


joewreschnig:
Giving up.

stefnotch:
@joewreschnig Does your implementation theoretically work? I'm willing to merge it. (I only became maintainer recently.)

joewreschnig:
I'm pretty drunk but I just re-read the linked paper (last paragraph second page) and it looks nearly right to me. Technically the code generates points [-1, 1) rather than (-1, 1) but I think the rejection deals with this case and even if it doesn't the effect should be negligible.

I don't know enough to say if it's still the most efficient way to generate such points.

stefnotch:
Could the rejection be avoided somehow? For example, with a bunch of extra calls to `Math.sqrt()`. Calling `Math.random()` (`GLMAT_RANDOM()`) is fairly expensive.

stefnotch:
Well, I'm going to merge it in a week or so. 👍

stefnotch:
Ok, "merged".

PR #169, opened 2015-07-29
replace GLMAT_ARRAY_TYPE w/ glMatrix.ARRAY_TYPE

nickdesaulniers:
I think this was meant to be removed earlier.  Unfortunately, there still seems to be a reference to `GLMAT_ARRAY_TYPE` [here](https://github.com/toji/gl-matrix/blob/master/tasks/test/coverage.rake#L80), but I'm not sure if it's safe to remove or even what it's doing.


nickdesaulniers:
Also, I committed the modified build of dist/, let me know if you want me to remove that from this PR.

toji:
Looks good to me. The other reference you pointed out is from the old ruby-based build system. That's no longer in use, I just haven't pulled it out yet.

nickdesaulniers:
cool, filed #170

PR #172, opened 2015-07-29
add SIMD support to mat4.multiply

nickdesaulniers:
work in progress. let's discuss things such as:
* ~~correct runtime detection~~
* ~~how to test this [properly]? (I had an `|| true` and the SIMD.js polyfill)~~
* ~~garbage creation~~
* ~~should the loop be unrolled? or should we roll it up~~
* maturity of SIMD.js spec (might be a while)
* ~~profiling (which I've yet to do)~~
* 'SISD'?

or anything else

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
   "devDependencies": {
     "jasmine-node": "1.2.2",
     "node-libs-browser": "^0.5.2",
-    "webpack": "^1.9.10"
+    "webpack": "^1.9.10",
+    "simd": "2.0.0"
   },
   "license": "MIT"
 }
--- a/spec/gl-matrix/mat4-spec.js
+++ b/spec/gl-matrix/mat4-spec.js
@@ -19,8 +19,14 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE. */
 
 describe("mat4", function() {
+    var glMatrix = require("../../src/gl-matrix/common.js");
     var mat4 = require("../../src/gl-matrix/mat4.js");
     var vec3 = require("../../src/gl-matrix/vec3.js");
+
+    // Inject the polyfill for testing
+    if (!glMatrix.SIMD_AVAILABLE) {
+        require("simd").shim();
+    }
 
     var out, matA, matB, identity, result;
 
@@ -262,6 +268,21 @@ describe("mat4", function() {
                     0, 1, 0, 0,
                     0, 0, 1, 0,
                     1, 2, 3, 1
+                ]);
+            });
+        });
+        describe("when using SIMD", function() {
+            beforeEach(function () {
+                result = Array.prototype.slice.call(mat4.SIMD.multiply(
+                  new Float32Array(out), new Float32Array(matA),
+                  new Float32Array(matB)));
+            });
+            it("should work", function() {
+                expect(result).toBeEqualish([
+                    1, 0, 0, 0,
+                    0, 1, 0, 0,
+                    0, 0, 1, 0,
+                    5, 7, 9, 1
                 ]);
             });
         });
--- a/src/gl-matrix/common.js
+++ b/src/gl-matrix/common.js
@@ -28,6 +28,8 @@ var glMatrix = {};
 glMatrix.EPSILON = 0.000001;
 glMatrix.ARRAY_TYPE = (typeof Float32Array !== 'undefined') ? Float32Array : Array;
 glMatrix.RANDOM = Math.random;
+glMatrix.SIMD_AVAILABLE = (glMatrix.ARRAY_TYPE !== Array) && ('SIMD' in this);
+glMatrix.ENABLE_SIMD = false;
 
 /**
  * Sets the type of array used when creating new vectors and matrices
--- a/src/gl-matrix/mat4.js
+++ b/src/gl-matrix/mat4.js
@@ -24,7 +24,10 @@ var glMatrix = require("./common.js");
  * @class 4x4 Matrix
  * @name mat4
  */
-var mat4 = {};
+var mat4 = {
+  SISD: {},
+  SIMD: {},
+};
 
 /**
  * Creates a new identity mat4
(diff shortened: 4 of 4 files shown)


toji:
The code here looks really nice, and is generally what I'd want to see from this kind of optimization. I'd still love to understand the performance characteristics a bit better. I tried to do a quick and dirty jsperf comparison between the methods (http://jsperf.com/simdmat4), but jsperf is rather dead right now. :( I'll try and put together a local profiling script later.

Questions about detection and testing are good ones and I'll have to think on it a bit more. I'll definitely go on record saying glMatrix is in favor of loop unrolling, though!

For my information, what's the best way to get a browser (any flavor) that supports SIMD.js at the moment? It seems that the Firefox builds may only use it in the context of ASM.js?

sajjadt:
Currently, Firefox nightly build supports SIMD.js. I have some numbers to report from my experiments with SIMD version of Mat4 operations; the results are averaged over 10 iterations, each repeating the operation for 500,000 times. 

Multiplication
glMatrix - Avg: 20.2ms, Min: 19ms, Max: 21ms
glMatrix-SIMD - Avg: 6.8ms, Min: 5ms, Max: 15ms

Invert
glMatrix - Avg: 23.5ms, Min: 22ms, Max: 26ms
glMatrix-SIMD - Avg: 11.9ms, Min: 9ms, Max: 33ms

Translation
glMatrix - Avg: 4.8ms, Min: 4ms, Max: 6ms
glMatrix-SIMD - Avg: 1.9ms, Min: 1ms, Max: 4ms

Scaling
glMatrix - Avg: 5ms, Min: 4ms, Max: 7ms
glMatrix-SIMD - Avg: 1.6ms, Min: 1ms, Max: 3ms

Rotation (X axis)
glMatrix - Avg: 4.2ms, Min: 4ms, Max: 5ms
glMatrix-SIMD - Avg: 2.3ms, Min: 2ms, Max: 5ms

Rotation (Y axis)
glMatrix - Avg: 4.1ms, Min: 3ms, Max: 5ms
glMatrix-SIMD - Avg: 2.4ms, Min: 2ms, Max: 5ms

Transpose
glMatrix - Avg: 2.7ms, Min: 2ms, Max: 4ms
glMatrix-SIMD - Avg: 2ms, Min: 1ms, Max: 5ms

toji:
Those are really great numbers overall! Especially nice to see Multiplication approaching a 4X improvement on average.

Do you have any idea of what the garbage creation situation is like? I'd love to start working this into the code base, but if the tradeoff is XX more garbage collection pauses for YY more performance it will have to be considered carefully. Unfortunately that's the kind of behavior that averages like this tend to gloss over.

nickdesaulniers:
@sunfishcode has some insights.  Particularly that SIMD vectors are value types as opposed to object types.  Not sure if that makes a difference to the GC or not.

sunfishcode:
Once the JIT warms up, SIMD values are compiled to raw SIMD values in hardware registers rather than on the GC heap.

toji:
That makes a lot of sense, thanks or the clarification! Given that I don't see any reason not to move forward with this PR as long as the detection is working as intended and we're cone up with a testing pattern that we're comfortable with.

I'm also going to suggest that since this is technically a backwards-compatibility-breaking change we default SIMD behavior to disabled during early development so that it doesn't disrupt current users but allows adventurous developers to test and contribute.

nickdesaulniers:
> as long as the detection is working as intended

I'd like another day to do more testing. Particularly, I'm curious if various build systems might redefine `this` to not refer to the global context.

> this is technically a backwards-compatibility-breaking

Is it?  I can't spot where.

> default SIMD behavior to disabled during early development

Yeah, I think that's a good idea.  No rush on enabling SIMD fast paths, or even this pull request.  Please invite others for their input.  Particularly, I'd like to see if there's a way to "design for test," as in to see if we can inject the SIMD polyfill to test for correctness.  Maybe I can whip that up in a separate branch and we can land test harness support before this?

toji:
One of the more common patterns with glMatrix (which may even be used in
the tests, I don't recall) is to use online arrays as inputs. Like so:

mat4.translate(out, in, [10, 0, 5]);

If that's using SIMD on the back end that shorthand breaks. Same goes for
if you're using arrays-as-vectors from some other third party library. The
strict requirement for TypedArrays will probably cause a lot of minor
breakage across the board. Most of it should be easily fixable but it's
still a breaking change.

It's worth noting that glMatrix was first built at a time when TypedArrays
were still a poorly supported and WebGL-specific concept. We've largely
moved past that now, so I have no qualms about gently guiding the library
towards TypedArray use exclusively. Maybe that's a good target for glMatrix
3.0? TypedArrays everywhere with a focus on SIMD? In the meantime having it
optional in 2.x provides a good migration/experimentation path.

nickdesaulniers:
That sounds like a good plan; thanks for pointing out the compatibility implications.  Shall we disable for now with an `&& false` or have two distinct flags `SIMD_AVAILABLE` and `DISABLE_SIMD`?

toji:
Yeah, let's do that. Minor style preference: ENABLE_SIMD instead.

nickdesaulniers:
I just had a thought on how we could test this.  We could reuse the idea of aliasing that's already used for `matX.mul`, just creating a `mat4.multiply` alias dynamically.  So instead of:

```coffeescript
if simd
  return anonymous_function
else
  return anonymous_function
```
we'd have something more like:
```coffeescript
mat4._SIMDmultiply = ->
  ...
mat4._nonSIMDmultiply = ->
  ...
mat.multiply = (->
  if simd
    return mat4._SIMDmultiply
  else
    return mat4._nonSIMDmultiply
)()
```
That way, in the tests for mat4, we can explicitly test both `mat4._SIMDmultiply` and `mat4._nonSIMDmultiply`. The downside here is that we would be exposing the two underscore-prefixed methods.  What are your thoughts on this idea?

toji:
I'm not opposed at all to exposing the two function variants, and in fact I don't think we should try and "hide" them either. What if we simply did:

```mat4.multiplySIMD = ...
mat4.multiply<Something?> = ...
mat4.multiply = (
  if simd
    return mat4.multiplySIMD
  else
    return mat4.multiply<Something?>
)()```

or alternately (and I think I like this better):

```mat4.SIMD.multiply = ...
mat4.<Something>.multiply = ...
mat4.multiply = (
 // You get the idea
)()```

Then testing is easy, and it benefits people trying to do code migration. They can switch on SIMD at some point and then replace a few incompatible bits with the explicitly backwards compatible version, or gradually replace bits of their code base with the SIMD version instead. Sounds like a win-win to me!

The hard part, of course, is figuring out the name for the non-SIMD path. :) "Legacy" sounds premature, "Compatibility" is too lengthy to expect anyone to type, "Core" makes it sounds like SIMD is a super set, as does "Base" or "Basic", "Array" is confusing because both SIMD and non will use TypedArrays. Hm...

Bikeshedding welcome on this one. ;D

**Comment by pastasfuture** [comment]
How about "SISD" :P

**Comment by pastasfuture** [comment]
Been :ghost: lurking on this thread. Looking forward to this.

**Comment by toji** [comment]
"SISD" is actually not terrible, but I'm concerned that it's maybe a bit obscure for the average web developer who hasn't had to muck around in the world of SIMD until now. It also wouldn't hurt if it was a bit more visually distinct from "SIMD" at-a-glance to help with code readability. Not gonna rule it out just yet, though.

**Comment by pastasfuture** [comment]
Agreed, quick readability and autocomplete are pretty important here.

**Comment by acterhd** [comment]
I also made own SIMD.

**Comment by nickdesaulniers** [comment]
Ok, so it looks like the polyfill is not an npm module.  This patch is not mergeable as is, since I did not commit the pollyfill itself.  Should we ask [them](https://github.com/tc39/ecmascript_simd/issues) to publish to npm and add it to our devDependencies?  Also, should I reimplement all of the tests from SISD for SIMD? [my guess would be yes].  Also, still open to ideas for replacing 'SISD'.  Maybe 'scalar'?

**Comment by PeterJensen** [comment]
@sajjadt and others.  We really appreciate this (on behalf of the SIMD.js team).  I was curious about the numbers you reported for the mat4.invert() improvement.  A while ago I wrote an invert benchmark:

https://github.com/tc39/ecmascript_simd/blob/master/src/benchmarks/inverse4x4.js

It's using Cramer's rule.  I saw close to 8x speedups.

Have you posted your source code anywhere?

**Comment by littledan** [comment]
I'm really happy to see this kind of usage! Do you have any feedback for the SIMD.js spec? It's now right around the end of when it'll be OK to make changes.

**Comment by nickdesaulniers** [comment]
I pester @sunfishcode all the time. ;)  I was asking for horizontal add for single precision floats, but that would only be useful to my initial naive implementation, so I don't really need that anymore.

IRT this particular PR, we now have left to decide on `SISD` or w/e.  Also, @toji , should I make the tests now split based on SIMD/SISD?

```
describe 'when SISD'
  describe 'a'
  describe 'b'
  describe 'c'
  ...
describe 'when SIMD'
  describe 'a'
  describe 'b'
  describe 'c'
  ...
```

I would guess 'yes', but wanted to defer to the maintainer (not sure how attached to git blame you might be).

Also, @juj points out that there is indeed a faster implementation of mat4 multiply in his [MathGeoLib](https://github.com/juj/MathGeoLib/blob/master/src/Math/float4x4_sse.h#L306-L312).  I'm happy to translate that, but would ask @toji if he would consider it in a follow up PR.  Landing this implementation would free up others to begin additional implementations of other functions, since here we've mixed mat4 multiply, the runtime detection, and testing in one PR.

**Comment by PeterJensen** [comment]
@sajjadt I meant the source code for your SIMD version of mat4.invert(), not the mul()

**Comment by toji** [comment]
Yes on splitting the tests into SISD/SIMD, but as I think about it we should make every effort to not actually repeat tests. The code should produce the same results in either mode, with maybe some minor additions around testing what types are accepted. I'm no jasmine expert, but I believe that we should be able to pull off a pattern like this:

```
function mat4Tests(use_simd) {
  glMatrix.ENABLE_SIMD = true;
  return function() {
    var mat4 = require("../../src/gl-matrix/mat4.js");

    describe("create", function() { /* ... */ });
    // etc...
  };
}

describe("mat4 (SISD)",  mat4Tests(false));
describe("mat4 (SIMD)",  mat4Tests(true));
```

Of course that will be a decently large refactoring, so I'm not going to block the pull request on it. For now it's sufficient to just add new tests specifically for the new SIMD methods and we can make it pretty later. I'm also fine pulling the existing algorithm and then improving on it with a follow up PR. I think it's more important to start establishing the code patterns for the new paths than it is to get an optimal implementation on the first pass.

mhaghigh:
Folks, great to see this development! 
@sajjadt, excellent numbers. Looking forward to seeing SIMD.js use in CV libraries fro WebRTC.

sajjadt:
@PeterJensen I ended up using that. Here's the [link](https://github.com/sajjadt/gl-matrix/blob/master/src/gl-matrix/mat4.js#L1324) to the source code.

sajjadt:
@mhaghigh  That would be a great contribution for OpenCV. Did @huningxin work on Emscripten SIMD support?

toji:
This looks like it's in a pretty good state now. Still not sure if I'm happy with "SISD" long term, but that's a pretty easy search and replace if we decide to change it later.

Anything else you wanted to address before the first merge?

nickdesaulniers:
Go for it, follow up issues have been filed and are being tracked.  All systems go!  May a thousand SIMD flowers bloom.

toji:
BOOM! SIMD in da house!

nickdesaulniers:
:rocket: :rocket: :rocket:

benaadams:
:+1:

huningxin:
@nickdesaulniers , thanks for doing this. Really nice work!

> Did @huningxin work on Emscripten SIMD support?

I did. @juj is leading this effort. Please find more details at https://github.com/tc39/ecmascript_simd/issues/59#issuecomment-113824806

toji:
Quick note: After giving it some thought I decided to go with <type>.scalar as the alternative to <type>.SIMD. SISD was a good and as accurate as it gets, but being one character off from the alternative was bad for readability I felt. 'scalar' is still well understood in this context and more visually differentiated (and easier to autocomplete for you IDE types out there).

PR #173, opened 2015-07-31
Remove old Rake build system Fixes #170

### LINKED ISSUE #170: remove unused ruby build system

nickdesaulniers:
is it safe to remove all of this?  I'm not sure if there's still code coverage testing and version/tagging otherwise?


toji:
Yes, it's safe to remove. The newer system could still use a bit more polish, but you can still test and build easily with the node-based system, just not as a single command (yet). See TESTING.md and BUILDING.md.

PR #174, opened 2015-08-05
Will rework

acterhd:


nickdesaulniers:
I would recommend waiting until #172 lands, then refactoring this patch to work with the runtime detection and testing we've set up there.

acterhd:
Please add vec4.transformMat4

PR #180, opened 2015-08-07
Optimized SIMD mul function

sajjadt:
Ma4.mul is updated to replace Float32x4.Splat operations with Float32x4.Swizzle. Based on tests, it leads to about 10% performance improvement. Also, It's changed a bit for the sake of readability.


toji:
To be honest: I'm reading this on a phone in bed and the SIMD code it a little dense for me to wrap my head around right now on this little screen. So I'm just gonna take your word for it that this is faster and more readable. :) Thanks for the contribution!

nickdesaulniers:
@juj pointed out to me that this method is preferred; load + swizzle is faster than splat since load + swizzle are vector operations while splat is a scalar.

dist/gl-matrix.js
```diff
@@ -1697,51 +1697,53 @@ return /******/ (function(modules) { // webpackBootstrap
...
-	    t2 = SIMD.Float32x4.add(SIMD.Float32x4.mul(_d, t1), t2);
-	    SIMD.Float32x4.store(out, 12, t2);
-
+	    var a0 = SIMD.Float32x4.load(a, 0);
+	    var a1 = SIMD.Float32x4.load(a, 4);
+	    var a2 = SIMD.Float32x4.load(a, 8);
+	    var a3 = SIMD.Float32x4.load(a, 12);
+
+	    var b0 = SIMD.Float32x4.load(b, 0);
+	    var b1 = SIMD.Float32x4.load(b, 4);
```
juj: Vars `b1`-`b3` should not be loaded here, since if the browser executes this in the current form without intelligent optimizations, this causes register pressure for 12 simultaneous SIMD registers. It is best to organize the code strictly in the order of expected memory loads and stores, and use exactly the amount of SIMD variables that the register allocator should use registers. This eases the browser's register allocator work and avoids assuming that it would also be smart enough to reorder statements around to minimize register pressure.

The best order here is to load `a0`-`a3` first, then load `b0`, then compute `var out0 = ...` and store `out0`, then load `b1`, compute `out1` (to same variable as `out0`) and store it, and so on. That has register pressure of only 6 registers in total and the browser doesn't have to recognize statement reordering to optimize.

If one does have the 12 simultaneous registers to spend (x64 has 16 xmm registers available, x86 only 8, but it's a bit more complicated since more registers can spill caller's registers) and optimizing for fewer registers is not needed, then the computation of `out0` - `out3` can be split down to be done in parallel, which allows manually scheduling muls, adds and swizzles to pair up in separate execution ports.

nickdesaulniers: Yeah, did I mention that Jukka is a time traveler sent from the future to save us?

sajjadt: Good point, :+1:

PR #183, opened 2015-08-24
SIMD implementation of mat4.{scale, translate, transpose, adjoint, invert}

sajjadt:
Added SIMD implementation of mat4.{scale, translate, transpose, adjoint, invert} operations.


toji:
This is looking great! The only thing missing is tests, as mentioned in the previous patch. All you should need to do for that is to go into spec/gl-matrix/mat4-spec.js and change the calls for these functions from `mat4.scale` to `mat4.TEST_PATH.scale`, etc. That will make sure the tests exercise both paths.

sajjadt:
Thanks for the review! Tests now are updated to cover SIMD versions as well.

toji:
Perfect! Thanks for the contribution!

nickdesaulniers:
:beers: :rocket: :snowboarder:

PR #193, opened 2015-11-30
Implement add/subtract for mat2, mat2d, mat3, and mat4

thomcc:
I only needed this for mat3 but added it for the rest for consistency. It's certainly not as commonly needed as matrix multiplication, but it comes up frequently in physics problems.

I only did a scalar implementation for mat4, largely since SIMD is not well supported at all and so it's not something that is worth doing for me.

I'd also like to add matrix equivalents to `vec{N}.addAndScale`, (e.g. taking a scalar and another matrix, and working equivalently to `A = A + B*c` where A and B are matrices, and c is a scalar), since this is another operation I find myself doing. I didn't do this because I felt it would cause confusion with `mat{N}.scale` (for that matter, I'd also like to add multiplication of a matrix by a scalar...). If you can come up with better names for those, I'd be willing to implement them.

### DIFF
--- a/spec/gl-matrix/mat2-spec.js
+++ b/spec/gl-matrix/mat2-spec.js
@@ -207,4 +207,59 @@ describe("mat2", function() {
         });
    });
 
+    describe("add", function() {
+        describe("with a separate output matrix", function() {
+            beforeEach(function() { result = mat2.add(out, matA, matB); });
+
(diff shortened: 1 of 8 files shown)


toji:
Thanks for the additions, and specifically for including unit tests for them!

PR #194, opened 2015-12-01
Implement set and fromValues for mat2, mat2d, mat3, and mat4

thomcc:
Was surprised to find out that these did not already exist.

Let me know if there are any problems.


toji:
LGTM, thanks! And thanks for giving me something to test the new Travis CI config with! :D

PR #196, opened 2015-12-02
Implement equals function for all types

thomcc:
Strict equality comparison, e.g. ===. So this might do the wrong thing for NaNs, depending on your perspective.  

I considered adding a matching approxEquals as well, but considering that I didn't need it, and that the way to do accurately has some overhead that a lot of people might find undesirable (the obvious `Math.abs(x-y) < epsilon` fails for large x and y, so you need to do  `Math.abs(x-y) < epsilon*Math.max(1.0, Math.abs(x), Math.abs(y))`)... I punted on it.

### DIFF
--- a/spec/gl-matrix/common-spec.js
+++ b/spec/gl-matrix/common-spec.js
@@ -11,4 +11,17 @@ describe("glMatrix", function(){
     beforeEach(function(){ result = glMatrix.toRadian(180); });
     it("should return a value of 3.141592654(Math.PI)", function(){ expect(result).toBeEqualish(Math.PI); });
   });
+
+  describe("equals", function() {
+    var r0, r1, r2;
+    beforeEach(function() {
+      r0 = glMatrix.equals(1.0, 0.0);
+      r1 = glMatrix.equals(1.0, 1.0);
+      r2 = glMatrix.equals(1.0+glMatrix.EPSILON/2, 1.0);
+    });
+    it("should return false for different numbers", function() { expect(r0).toBe(false); });
+    it("should return true for the same number", function() { expect(r1).toBe(true); });
+    it("should return true for numbers that are close", function() { expect(r2).toBe(true); });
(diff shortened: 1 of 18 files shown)


toji:
Sorry about the conflicts, probably wouldn't have happened if I had merged in order. This is good functionality, but I'm trying to decide if it will be more confusing than helpful. In my experience vast swaths of developers don't grok the imprecision inherent in floating point math, and would expect this to work on the results of some computation. (Like "if a * b * c === identity").

It's hard to argue that a feature should be kept out simply because it might be misunderstood, but maybe the name could be changed to better indicate it's function? exactlyEquals? isIdentical? reallyTruelyEqualsThisIsProbablyNotTheFunctionYouAreLookingFor? ;)

thomcc:
I could rename this to exactEquals, and add an equals() that works with an epsilon if that would help. Only question is where to get epsilon. Obvious choices are glMatrix.EPSILON or an argument. Probably the former?

toji:
exactEquals sounds fine, and for consistency we should use glMatrix.EPSILON. I think that will handle 99% of people's needs.

thomcc:
Added both. Also added an approximate equality comparison (named `equals` for consistency) to glMatrix. 

The comparison method is different from the one you use in the testing code (which is an `abs(a-b) < epsilon` check), but it helps out a great deal for numeric robustness. You can read about this [here](http://realtimecollisiondetection.net/pubs/Tolerances/). 

But, it might be unintuitive, and has a small performance cost associated with it, so I'd be willing to replace it with an `abs(a-b) < epsilon` check if you think that would be better.

toji:
Looks good. I'm not super concerned about the performance here, as this is the type of function that's more likely to be used during testing or other non-perf sensitive code. If we start hearing complaints it's worth re-evaluating but I don't see that happening. Thanks for your continued contributions!

Issue #184, opened 2015-09-05
Add/Mul constants


jdavidberger:
I've noticed, somewhat surprisingly, there aren't functions to multiply / add scalars with vectors or matrices. This is somewhat common set of operations; any reason for this? Would a pull request for this addition be merged in?

stefnotch:
@jdavidberger Has this issue been addressed by the PR?

stefnotch:
I'll assume that it has been addressed and will close this issue.

PR #197, opened 2015-12-02
Add multiplyScalar and multiplyScalarAndAdd functions for matrices

thomcc:
Basically the matrix versions of `vec{N}.scale` and `vec{N}.scaleAndAdd` respectively. The name difference is due to the fact that `mat{N}.scale` is already taken.


thomcc:
This might address some of what #184 is asking for, but I'm not certain.

toji:
Not sure this entirely addressed #184, and honestly the multiplyScalarAndAdd seems kinda special case to me, but I'm happy to merge regardless.

PR #202, opened 2016-02-03
Include src folder in bower package

Manuel-S:
Hey, great library.
It would be nice if you would include the src folder in the bower package, that way I can load (and bundle) the modules I need directly in my application.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,6 @@
     "**/.*",
     "jsdoc-template",
     "spec",
-    "src",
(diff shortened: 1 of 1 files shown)


toji:
Sounds good to me.

Manuel-S:
thank you :+1:

Manuel-S:
Change doesnt seem to have appeared in the main bower repo yet, even though I cleared the cache I still only get dist.

Issue #136, opened 2014-09-22
Add support for matrix decompose and compose


sanjayc77:
Based on CSS3 transform spec: 

http://www.w3.org/TR/css3-transforms/#decomposing-a-3d-matrix
It also describes decomposing 2d matrix.

While we are at it, maybe support composing matrices given its component parts (translation, rotation, skew, perspective, etc).

Chromium also has an implementation in C++: 
https://github.com/ChromiumWebApps/chromium/blob/master/ui/gfx/transform_util.cc

stefnotch:
These function have been implemented.

gonnavis:
@stefnotch Hello, thank you very much for your work!

I saw `decompose` have been implemented, but `compose` not yet?

stefnotch:
@gonnavis See https://glmatrix.net/docs/module-mat4.html#.fromRotationTranslationScale and other matrix functions. To compose them, you multiply them.

The only thing that I think isn't in there is `skew`. There was [a pull request](https://github.com/toji/gl-matrix/pull/293) to add it a while back, but nothing happened since then. I suppose skewing is a quite rare operation.

gonnavis:
Thanks a lot!

PR #204, opened 2016-02-23
Add decomposition methods to mat4 and quat

sessamekesh:
**OVERVIEW**

I added methods for extracting useful information from mat4 and quat objects. This addresses issue https://github.com/toji/gl-matrix/issues/136 (which is also an issue I faced using gl-matrix in my own work).

For mat4, I added a mat4.getTranslation and mat4.getRotation method. They return the vec3 translation and quat rotation stored in a transformation matrix, respectively.

For quat, I added a quat.getAxisAngle method. This returns the unit vector axis and angle (radians) rotation stored by the input quaternion.

**MATHEMATICAL CONSIDERATIONS**

A single axis/angle rotation may be stored by at least two different quaternions. If an axis of rotation is [a, b, c] and the angle of rotation d, the same rotation may be described by an axis of rotation [-a, -b, -c] and an angle 2*PI - 2. The axis and angle returned by quat.getAxisAngle are not guaranteed to be the same that were provided to a quat.setAxisAngle call, but they are guaranteed to represent the same rotation.

A quaternion storing no rotation will always return the axis [1, 0, 0] and angle 0, arbitrarily - the axis itself is not important in this case.

Decomposing a quaternion into its axis and angle components has a potential to yield floating point error. This also applies to decomposing a transformation matrix into its rotational component. Any who use this method should be aware of this limitation. During testing with random values, I came across several cases which were off a relatively large degree of error for 64 bit floating point values after repeated composing and decomposing.

**POTENTIAL ISSUES**

I'm not extremely familiar with this project, so these are things I was unsure about. If you prefer a different style, let me know and I can make the appropriate changes.

quat.getAxisAngle cannot mutate a passed in Number parameter by reference. Currently, I have the method returning the angle and mutating the axis array passed in the parameter list. Of several less than ideal alternatives, I thought this the most elegant, but it doesn't fit the style of the rest of the code.

Unit tests for mat4.getRotation use more newlines than many other unit tests in the file. This was done for clarity, though if brevity is preferred I can easily make those changes.

**EXTERNAL SOURCES USED**
Extract rotation quaternion from 3x3 matrix: http://www.euclideanspace.com/maths/geometry/rotations/conversions/matrixToQuaternion/index.htm

### DIFF
--- a/src/gl-matrix/quat.js
+++ b/src/gl-matrix/quat.js
@@ -193,6 +193,35 @@ quat.setAxisAngle = function(out, axis, rad) {
 };
 
 /**
+ * Gets the rotation axis and angle for a given
+ *  quaternion. If a quaternion is created with
+ *  setAxisAngle, this method will return the same
+ *  values as providied in the original parameter list
+ *  OR functionally equivalent values.
+ * Example: The quaternion formed by axis [0, 0, 1] and
+ *  angle -90 is the same as the quaternion formed by
+ *  [0, 0, 1] and 270. This method favors the latter.
+ * @param  {quat} q     Quaternion to be decomposed
+ * @param  {vec3} axis  Vector receiving the axis of rotation
+ * @return {Number}     Angle, in radians, of the rotation
+ */
+quat.getAxisAngle = function(q, axis) {
+    var rad = Math.acos(q[3]) * 2.0;
+    var s = Math.sin(rad / 2.0);
+    if (s != 0.0) {
+        axis[0] = q[0] / s;
+        axis[1] = q[1] / s;
+        axis[2] = q[2] / s;
+    } else {
+        // If s is zero, return any axis (no rotation - axis does not matter)
+        axis[0] = 1;
+        axis[1] = 0;
+        axis[2] = 0;
+    }
+    return rad;
+};
+
+/**
  * Adds two quat's
  *
  * @param {quat} out the receiving quaternion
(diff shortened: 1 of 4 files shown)


src/gl-matrix/quat.js
```diff
@@ -193,6 +193,35 @@ quat.setAxisAngle = function(out, axis, rad) {
...
+ *  values as providied in the original parameter list
+ *  OR functionally equivalent values.
+ * Example: The quaternion formed by axis [0, 0, 1] and
+ *  angle -90 is the same as the quaternion formed by
+ *  [0, 0, 1] and 270. This method favors the latter.
+ * @param  {quat} q     Quaternion to be decomposed
+ * @param  {vec3} axis  Vector receiving the axis of rotation
+ * @return {Number}     Angle, in radians, of the rotation
+ */
+quat.getAxisAngle = function(q, axis) {
```
toji: I think that for the sake of consistency this should be: `function(out_axis, q)`. I'll admit it reads a little funny, but every other function in the library has the output vector/mat/quat as the first argument so we probably shouldn't break that pattern here.

toji:
This is really excellent, high quality work! Thank you! (And especially thank you for providing unit tests up front!)

I've left one comment about a minor style change, but once that's fixed I'm happy to merge!

sessamekesh:
I added the requested style change - I'm glad everything else lines up well. Thanks for the compliment, and thanks for the project - it has been really useful!

toji:
Awesome, thanks!

sessamekesh:
There's no _good_ reason I didn't include a `getScale` method - my particular use case called for translation and rotation only. If I were to do it, I would probably just do the math directly on the individual cells instead of pulling columns from the matrices. AFAIK there is no built in method to pull out the columns of a matrix, but I also haven't searched through recently.

PR #205, opened 2016-02-24
Implement vec{2,3,4}.{ceil,floor,round}

pyrotechnick:
I've been monkey-patching these into `gl-matrix` for years; it's time they're merged in.

They're particularly useful when working with 2D grids, 3D lattices and bounding volumes.

The vec4 variants are useful for ray-casting in 4D; where a surface exists only for an interval in time. 

I've attempted to logically sequence the methods within the appropriate files.

I've also provided tests utilising constants from the `Math` object.

Let me know if you require any adjustments before merging.


toji:
Looks good. Thanks for providing the unit tests!

PR #209, opened 2016-03-29
[WIP] ES6 module conversion

bjornharrtell:
For fun I started an effort to migrate the code to use native ES6 modules.

ES6 modules can replace the use of CommonJS modules. It can be done without breaking the ability to use a module bundler to produce a compatible single file build similar to the the current `dist/gl-matrix.js` using the same entry point `src/gl-matrix.js`. The benefit is that other projects that also use ES6 modules can import parts of gl-matrix statically (entry point not used) and in that case only what is selectively imported will be included the the application build (assuming a bundler that can do "tree-shaking").

If there is interest I will try to finish this work. Remaining tasks: 
- [x] Convert remaining source files
- [x] API compatibility (AFAIK)
- [x] Tweak test suite to work with ES6 modules
- [x] Do not break the rule that ES module namespace exotic objects are not extensible.
- [ ] Make sure `typeof SIMD !== 'undefined'` works as a replacement for `'SIMD' in this`
- [ ] Reduce changes to a minimum
- [ ] Make sure changes after f228202d609e2244dd1c0f881fa026d748922e19 have been merged correctly


toji:
I'm definitely interested in seeing this happen! I don't think I'd want to merge until the full codebase has been converted, but I think it's a great step for the library to take.

bjornharrtell:
Nice @toji :) I've worked towards full conversion. As you can see tests pass at Travis, but there is one 1 pending case not sure what that could be. Still some remaining tasks to be done, I've added them to the list in the issue description.

bjornharrtell:
I've moved mat4 SIMD implementation to a separate module, but perhaps that was a bit hasty...

toji:
Sorry being slow to get back on this. I'm excited for this update, but have also been super busy. :P I'll do a more thorough review soon, but I did want to say that I'd prefer that the SIMD and Scalar implementations stay part of the same file. There's been a lot of thought put in to how to allow developers to cleanly upgrade their code base once SIMD is more widely available, and we generally don't want them to have to change individual calls or include anything differently. Should just be a bool that gets flipped.

Still, this is fantastic work! Thanks for putting the time into the conversion!

bjornharrtell:
No problem I'm super busy too. :P

I see two issues here - one is avoiding API breakage, the other is the ability to switch to SIMD dynamically or not. About the API I'm not sure what is public and and not in the old code base. I assumed `mat4.transpose` is public API and that `mat4.SIMD.transpose` is internal or private, since only `mat4.transpose` is tested. If my assumption is correct, moving SIMD-implementation does not affect the public API? In the case of enabling SIMD with a runtime boolean I can make it so but the drawback is that both non-SIMD and SIMD needs to be imported, since importing modules with the `import` statement is static. Also I was a bit confused about ENABLE_SIMD beeing set to false at https://github.com/toji/gl-matrix/blob/master/src/gl-matrix/common.js#L31 so it did not seem it could be changed dynamically in the first place?

toji:
So here's the tl;dr on SIMD integration. It's not dynamically togglable, and the `*.SIMD.*` and `*.scalar.*` function variants need to be publicly exposed. (It's also a work in progress)

SIMD requires that you're using TypedArrays, which is what glMatrix creates and uses by default, but it also allows you to pass in an indexable array. So `mat4.translate(out, in, [1, 2, 3]);` is totally valid right now but it will fail if the backing implementation is using SIMD. As such we're leaving the SIMD and scalar variants directly accessible so that we we do turn it on developers can easily pinpoint cases like the one above and replace it with `mat4.scalar.translate(out, in, [1, 2, 3]);` to get it working. Or, alternately, if their whole code base is like that but they want to optimize a hot function they can explicitly use `mat4.SIMD.translate(out, in, vec);` without converting everything. Relatedly `ENABLE_SIMD` isn't meant to be dynamic, because all it does it determine whether `mat4.translate` maps to the SIMD or scalar variant at start up time.

As for the testing, I know `mat4.multiply` handles testing of both paths. That should be looked at as the canonical example of how to test functions, and if any SIMD-converted functions don't yet do that it's a bug.

bjornharrtell:
Thanks, that makes things more clear to me.

The drawback here is that exposing both `mat4.SIMD` and `mat4.scalar` will, if I understand it correctly, disable the ability to "tree-shake" if using ES6 modules directly in an application which means that both `mat4.SIMD` and `mat4.scalar` will be included even if only one of them is used. This is however of course a much lesser issue than breaking API but perhaps it should be reevaluated for 3.x. Objects-as-namespaces is no longer a good practice, I think.

I'll try to find time to revise this PR to keep the `mat4.SIMD` and `mat4.scalar` as is. Perhaps also keep `mat4` in one file, but it is a bit large and the API doesn't have to be affected by a breakup. :)

toji:
For 3.x, since it's implied to be a breaking change, I'd like to enforce the SIMD route all the time. That kinda depends on browser support, though. I'm also happy to re-evaluate the naming if there's a more ES6-y pattern that should be used.

bjornharrtell:
So had another go at this and I hope to have achieved API compatibility for `mat4` as discussed. I've moved both SIMD and scalar implementation to separate files, importing them to `mat4` with conditional use and also exporting them to `mat4.SIMD` and `mat4.scalar` in the entry point. I've also adapted the current webpack build config to work instead of using rollup.

I think I should be able to reduce the changes introduced (like the whitespace changes in specs) and have added that as a remaining work item in the issue description. I've not done much testing outside the test suite, so obviously this PR need more serious real world testing but I'm ill equipped to do so.

src/gl-matrix.js
```diff
@@ -26,12 +26,38 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+import * as mat3 from "./gl-matrix/mat3.js";
+import * as mat4 from "./gl-matrix/mat4.js";
+import * as mat4scalar from "./gl-matrix/mat4scalar.js";
+import * as mat4simd from "./gl-matrix/mat4simd.js";
+import * as quat from "./gl-matrix/quat.js";
+import * as vec2 from "./gl-matrix/vec2.js";
+import * as vec3 from "./gl-matrix/vec3.js";
+import * as vec4 from "./gl-matrix/vec4.js";
+
+mat4.scalar = {
```
andyearnshaw: I believe this and line 52 would throw errors in a true ES module environment; ES module namespace exotic objects are [not extensible](http://www.ecma-international.org/ecma-262/6.0/#sec-module-namespace-exotic-objects).

bjornharrtell: Interesting. This is one part of ES modules that I have not fully understood yet but your link will be helpful.

bjornharrtell: Think I've got it now and have added a commit that I think is the correct way to do it.

andyearnshaw: Yeah the new commit looks good.  Thanks, and apologies for the abrupt drive-by comment before, I was in a bit of a rush.

toji:
Doing a more thorough review now. I'm a little wary of this breaking some people's use cases, but I'm also not sure how we'd find that out unless we update and wait for the cursing. :)

Do you know what the merge conflict is?

bjornharrtell:
Yep it's https://github.com/toji/gl-matrix/commit/33045e8e5e678e43983a244cb00e0776aed8f7d9, which is problematic in either case because `this` is not defined in a module. Same issue as with SIMD-detection. I've rebased the branch but the change from 33045e8e5e678e43983a244cb00e0776aed8f7d9 is not included. I'm not sure what the best solution is here and unfortunately I'm not even sure the capability detection works at all in this branch, it's untested.

andyearnshaw:
@bjornharrtell how's this coming along? I'm keen on progress here because this will be really useful with rollup's tree shaking.

maxdavidson:
@bjornharrtell Is the capability checking really a blocker? If you want to minimize the changes and not use the `typeof` idiom, you could just fetch the global object directly and keep the code essentially the same.
````javascript
const global = new Function('return this')();
export const SIMD_AVAILABLE = (ARRAY_TYPE === global.Float32Array) && ('SIMD' in global);
````

mourner:
@bjornharrtell FYI, we needed a slim build of gl-matrix, so I created a fork with ES6 modules + Rollup and without any SIMD code. Feel free to use: https://github.com/mapbox/gl-matrix

bjornharrtell:
Nicely done @mourner, and thanks for letting me know. :)

bjornharrtell:
Closing this as it's outdated and would need a reboot anyway.

Danielhu229:
What a pity

PR #211, opened 2016-04-04
Prevent Float32Array undefined error on missing typed array support.

aunzat:
Fix for missing typed array support in old browsers.


toji:
Thanks!

PR #230, opened 2016-10-31
fixed precision issues in vec3 angle function

mlimper:
Example:
var vec1 = [0.024432172998785973, 1.3685636623961273e-8, 0.9997014999389648]
var vec2 = [-0.024432172998785973, -8.114220406696404e-9, -0.9997014999389648]

dot(vec1, vec2) will be -1.0000000200576589

=> For values smaller than -1, the `Math.acos`function returns `NaN` (at least in my Chromium browser), therefore the additional check ensures that the cosine value from the dot product is clamped (as it was already done for values greater 1)


toji:
Thanks!

Issue #221, opened 2016-09-07
Dual Quaternions


stefnotch:
Would dual quaternions be a useful addition to this library? I would be willing to write the code for them.

Edits: 
Code is [here](https://github.com/stefnotch/gl-matrix)
Documentation: https://stefnotch.github.io/gl-matrix/

Todo list:
- `SLERP`/`ScLERP`
- `lookAt`
- `targetTo`
Not so important:
- `forEach`
- `calculateW`

chinedufn:
When dual quaternion blending all of the dual quaternion operation end up simplifying down to regular quaternion operations (AFAIK).

Is there some other application of dual quaternion that you're looking to support? Or were you thinking of a dual quaternion convenience wrapper around the more "basic" math?

stefnotch:
Well, Dual Quaternions are really useful when it comes to skeletal animation. 
Usually, you tend to have a fair number of bones, which need to be animated. That's rather performance heavy. So, what I am trying to do is to first get a basic wrapper done, then optimize it.

I already got a fair number of functions done.

chinedufn:
Which functions? Asking because I'm not sure if you need any dual quaternion functions for skeletal animation. The useful ones all simplify down to a combination of regular quaternion functions. So it seems like you'd just be writing convenience wrappers around quaternion math.

---

On the CPU side (thus anything that this library might optimize) of Dual Quaternion Linear Blending (skeletal animation) you're usually just linearly blending between two dual quaternion keyframes per joint based on the current time.

Which boils down to blending their rotation quaternions and the translation quaternions.

I'm curious about how you might optimize that?

And please let me know if I'm missing something!

chinedufn:
Although to be fair... It's probably much easier for someone to just grab a dual quaternion library than to learn the underlying quaternion math.

Curious to see what you come up with! Not sure if it would get merged into `gl-matrix` but might still have a nice home as its own module regardless

stefnotch:
First of all, you need to turn a quaternion/rotation and a translation into a dual quaternion. (That's one function which isn't that simple and can be optimized a bit.) 

Multiplying dual quaternions (pretty much the same as multiplying matrices) isn't that straightforward either. (I guess it could be optimized a bit. That optimization is quite important, since models tend to have a lot of bones and there can be a lot of models on the screen.)

Then, there also are the from/to matrix conversions. (Which can't be simplified down to quaternions.)

[...]

Issue #108, opened 2014-02-14
vec3.transformQuat uses slow implementation?


paulharris:
Hi,

in vec3.js, there is a URL pointing to benchmarks.
The benchmarks indicate that the OTHER way of implementing .transformQuat() is a lot faster.

Why are we using the slow version?

cheers

stefnotch:
Looks like the benchmark's other implementation is a broken one. The cross product wasn't implemented correctly.
Here is a fixed one: https://github.com/toji/gl-matrix/pull/231 (And also a dual quaternion implementation)

PR #231, opened 2016-11-10
quat2 implementation

stefnotch:
I implemented [dual quaternions](https://github.com/toji/gl-matrix/issues/221), which are extremely useful for gpu skinning. CryEngine uses them!
Dual quaternions are very similar to matrices. They represent a rotation and translation using 2 quaternions.
## Pros:
+They only use 8 floats (rotation, translation) instead of 16 floats.
+They do not drift over time and do not introduce unwanted scaling and shearing.
+Interpolating them is easier and they do not suffer from the [candy wrapper effect](http://libigl.github.io/libigl/tutorial/images/arm-dqs.jpg). The top one uses linear blend skinning with matrices and the bottom one uses dual quaternion skinning.

## Cons:
-They cannot represent shearing.
-GLSL does not support them out of the box. (You have to write the multiplication code yourself; it is almost as fast as a matrix multiplication.)

I also improved the [vec3.translateQuat](https://github.com/toji/gl-matrix/issues/108).
Here is a fixed benchmark: https://jsperf.com/quaternion-transform-vec3-implementations-fixed (The benchmark's code was wrong)

If you feel like this doesn't fit into the library, feel free to not accept the pull request.


stefnotch:
I finally finished optimizing all of the code. What do you think about it?

[Here](https://github.com/stefnotch/gl-matrix) is the code
Documentation: https://stefnotch.github.io/gl-matrix/

Should I remove some of the useless comments. (The previous implementations of a function...)
Should I remove the useless comments?

kevzettler:
Further discussion of this PR at: https://github.com/toji/gl-matrix/issues/221 

I think this is a valuable addition to the library. Was about to start implementing these functions myself and was glad to find this PR and issue

src/gl-matrix/mat4.js
```diff
@@ -946,6 +946,38 @@ export function fromRotationTranslation(out, q, v) {
   return out;
 }
 
+//TODO FIX ME
```
kevzettler: Whats the Issue here?

src/gl-matrix/vec3.js
```diff
@@ -520,31 +520,43 @@ export function transformMat3(out, a, m) {
   return out;
 }
 
+//TODO Fix me (upgrade the function signature)
```
kevzettler: TODO whats the issue with the signature?

kevzettler:
@stefnotch I left some comments on some `TODO FIX ME` lines. It's unclear to me if there are outstanding issue with the PR.

> Should I remove some of the useless comments. (The previous implementations of a function...)

It is a best practice to remove any dead comments, or code from a final PR.

stefnotch:
Great to see that somebody is interested! I'll go and fix those as soon as possible. (Should be done by next Sunday. Hopefully sooner.)

docs/quat2.js.html
```diff
@@ -0,0 +1,948 @@
...
+//TODO: quat2.forEach
+
+//LINKS: https://github.com/bobbens/libdq
+//https://github.com/markaren/DualQuaternion/tree/master/src/main/java/info/laht/dualquat
+//http://wscg.zcu.cz/wscg2012/short/A29-full.pdf
+//http://www.xbdev.net/misc_demos/demos/dual_quaternions_beyond/paper.pdf
+//https://github.com/mosra/magnum/blob/master/src/Magnum/Math/DualQuaternion.h#L364
+//http://mosra.cz/blog/magnum-doc/classMagnum_1_1Math_1_1DualQuaternion.html#a6dcad20c0e9c7c9361b9c291ee42b181
+/**
+ * @class Dual Quaternion&lt;br>Format: [[real],[dual]]&lt;br>Make sure to have normalized dual quaternions, otherwise the functions may not work as intended.
```
kevzettler: confused here isn't the format `[real, dual]` ?

docs/quat2.html
```diff
@@ -0,0 +1,168 @@
...
+
+
+
+<section>
+
+<header>
+    
+        <h2><span class="attribs"><span class="type-signature"></span></span>quat2<span class="signature">()</span><span class="type-signature"></span></h2>
+        
+            <div class="class-description">Dual Quaternion<br>Format: [[real],[dual]]<br>Make sure to have normalized dual quaternions, otherwise the functions may not work as intended.</div>
```
kevzettler: Should the format be `[real, dual]` here?

stefnotch:
Oh, thank you very much for pointing out those things. Looks like I totally missed them.

kevzettler:
@stefnotch I would just like to add that I started using your branch in a project and dropped in 
```
            var rotatedJoint = quat2.rotateY([],
                                         lowerInterpolatedJoints.joints[jointId],
                                         -player.moveDirectionVelocity[1] * (Math.PI/6))
```

and It worked as expected! 

I'm already getting value out of this so great job!

stefnotch:
Ok, I would work on it, however there is a rather big, open issue right now: https://github.com/toji/gl-matrix/issues/260

docs/quat2.html
stefnotch: Yep

docs/quat2.js.html
stefnotch: Yeah, you are right

kevzettler:
Cool thanks for the heads up I can take a look at the tests

kevzettler:
Will need conflicts resolved now that test runner is fixed

stefnotch:
Ok, here is what I still have to do:
- slerp
- rotateAroundAxis (unit tests only)

Should I implement those 2 functions?
- lookAt
- forEach

kevzettler:
great work on this!

I need some clarification on what you are asking.

> Should I implement those 2 functions?
> 
> lookAt
> forEach

Are you saying you haven't implemented these two functions? or wrote tests for them? If that's the case I think `lookAt` would be important to implement. Reasons being; its consistent with the other data structure functionality in `gl-matrix`.  

`forEach` I'm not as sure on. Is it an iterator method from the individual `quats` in a `quat2` ? 

Either way, I don't think either `lookAt` or `forEach` is a blocker to this PR. They can be added in on later PR's when there is community demand for it.

stefnotch:
I have not implemented these 2 functions yet. 

Ok, I guess I'll implement lookAt! :)

forEach would be similar to [vec3.forEach](http://glmatrix.net/docs/module-vec3.html#.forEach). I guess I won't implement it for now and add it at a later date if there is any demand for it.

stefnotch:
@kevzettler @toji @sinisterchipmunk @chinedufn 
I'm happy to report that I am done with the dual quaternion implementation!

docs/mat4.js.html
```diff
@@ -974,6 +974,38 @@ <h1 class="page-title">Source: mat4.js</h1>
   return out;
 }
 
+//TODO FIX ME
```
kevzettler: Is this outstanding?

docs/vec3.js.html
```diff
@@ -548,31 +548,43 @@ <h1 class="page-title">Source: vec3.js</h1>
   return out;
 }
 
+//TODO Fix me (upgrade the function signature)
```
kevzettler: Is this new? why does the signature need to change?

spec/gl-matrix/quat2-spec.js
```diff
@@ -0,0 +1,734 @@
...
+        outQuat = [0, 0, 0, 0];
+        vec = [1, 1, -1];
+        id = [0, 0, 0, 1, 0, 0, 0, 0];
+        deg90 = Math.PI / 2;
+    });
+
+    describe("translate", function() {
+        var matrixA = mat4.create(), matOut = mat4.create(), quatOut = quat2.create();
+        beforeEach(function() {
+              //quat2A only seems to work when created using this function?
```
kevzettler: Any idea whats up here?

spec/helpers/spec-helper.js
```diff
@@ -3,6 +3,11 @@ const assert = require('assert');
 
 
 global.expect = function(e) {
+    
+    function expected(e, message, a) {
```
kevzettler: I don't see this used anywhere below

src/gl-matrix/mat4.js
```diff
@@ -947,6 +947,37 @@ export function fromRotationTranslation(out, q, v) {
...
+ * Creates a new mat4 from a dual quat.
+ *
+ * @param {mat4} out Matrix
+ * @param {quat2} a Dual Quaternion
+ * @returns {mat4} mat4 receiving operation result
+ */
+export function fromQuat2(out, a) {
+    //var normalizedA = quat2.create();
+    //quat2.normalize(normalizedA, a);
+    //quat2.getTranslation(translation, normalizedA);
```
kevzettler: remove unused commented code

src/gl-matrix/vec3.js
```diff
@@ -522,29 +522,40 @@ export function transformMat3(out, a, m) {
...
 /**
  * Transforms the vec3 with a quat
+ * Can also be used for dual quaternions. (Multiply it with the real part)
  *
  * @param {vec3} out the receiving vector
  * @param {vec3} a the vector to transform
  * @param {quat} q quaternion to transform with
  * @returns {vec3} out
  */
 export function transformQuat(out, a, q) {
```
kevzettler: I don't think we should rewrite this function to handle parameter overriding for both `quat` and `quat2` types. The `gl-matrix` library aims towards a 'functional programming' style of 'pure functions' which aims for static type style fixed function signatures.  I would propose a new `transformQuat2` method that explicitly handles the `quat2` behavior.

spec/gl-matrix/vec2-spec.js
```diff
@@ -330,7 +330,7 @@ describe("vec2", function() {
 
         beforeEach(function() { result = vec2.distance(vecA, vecB); });
 
-        it("should return the distance", function() { expect(result).toBeCloseTo(2.828427); });
+        it("should return the distance", function() { expect(result).toBeEqualish(2.828427); });
```
kevzettler: Why the change on assertions here?

docs/mat4.js.html
stefnotch: That's a documentation file. I didn't rebuild the docs. (Should probably do it)

spec/gl-matrix/vec2-spec.js
stefnotch: I removed `toBeCloseTo` because it wasn't working properly and `toBeEqualish` does the exact same thing.

spec/helpers/spec-helper.js
stefnotch: It's a convenience function which is only used in the same file.
e.g. https://github.com/stefnotch/gl-matrix/blob/master/spec/helpers/spec-helper.js#L86

src/gl-matrix/vec3.js
stefnotch: :) I'm perfectly aware of it. 

This thing works because a dual quaternion is represented by a flat array with 2 quaternions. 
`[x0, y0, z0, w0,  x1, y1, z1, w1]` 
The first part is a perfectly normal, normalized quaternion which represents a rotation. The second part is a quaternion which represents a translation. 

And, the function only looks at the first 4 elements of whatever array gets passed to it. In our case, it's the dual quaternion's rotation. It ignores the translation part, which is perfectly fine, because [a vector doesn't really have a translation.](https://en.wikipedia.org/wiki/Euclidean_vector) (When you have a vector that represents a point, it's actually a vector from the origin to the point's coordinates.)

However, this comment might also cause some confusion. e.g.
> Why doesn't this translate my vector?
transformQuat(out, [0,0,0], [no rotation, a translation])

(The answer would be: Because a vector only has a "magnitude (or length) and direction" --Wikipedia)

So, I guess it can be removed? If a person actually wants to rotate his vectors with a dual quaternion's rotation part, he probably knows his stuff and doesn't need this comment?

kevzettler:
@stefnotch thanks for addressing comments. This lgtm, additional review from one of the maintainer's cc @toji @sinisterchipmunk is needed.

To re-emphasize the value prop here, I feel this is a relevant and powerful addition to the `gl-matrix` library. dual-quats are incredibly useful for skeletal animation. I personally am currently using dual-quat features with @chinedufn's https://github.com/chinedufn/skeletal-animation-system which is built on dual quats

stefnotch:
So far, over a month has passed by and nothing has happened.

kevzettler:
ping @toji @sinisterchipmunk

stefnotch:
Note to self: Make sure that the quat2 code doesn't mess up the other code.

stefnotch:
@chinedufn @kevzettler @Beeblerox

Dual quaternions are now a part of gl-matrix! 😃 Please report any bugs that you find. I'm also open to all suggestions.

PR #233, opened 2016-11-19
implement quat.fromEuler()

jwu:
Just add this interface to help people easily create a quaternion from the familiar euler system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 tmp
 node_modules
 npm-debug.log
+.vscode/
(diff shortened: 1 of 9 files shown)


jnsmalm:
Why has this not been merged? It would be useful.

jeromeetienne:
i would find it usefull too

toji:
#249 contained the `quat.fromEuler()` function, but not the vector inverts. Probably best to do those as a separate pull request.

PR #248, opened 2017-07-03
added mat3.projection

mreinstein:

Issue #244, opened 2017-05-16
Using without polluting the global namespace


computersarecool:
At risk of repeating #112 (which was closed without a solution), in the browser, when including this, all of teh classes are added to the global (window) object.

Would you guys be willing to encapsulate this more so we do not have to use something like browserify to keep this out of the window object?

I think that adding everything to a glMatrix object (which itself if a member of the window object) would be very helpful

toji:
#249 converts the library to use ES2015 modules. The pre-compiled scripts in the `dist/`directory will continue to place all functions in the global namespace, but libraries that want to include a more targeted set of functionality can now use the individual modules directly.

Issue #240, opened 2017-03-07
VERSION is not updated


laurennlam:
Hi all, just to inform that the file VERSION is not up to date. That can be confusing because I was referring to this file during my development. It takes me a few minutes to understand that it wasn't the right version :)
I can make a pull request if you prefer.

toji:
Fixed in #249

Issue #237, opened 2017-02-11
Small error in the documentation


ArnaudDroxler:
In the documentation page of the vec3, there's a small error.
It's about the fonction transformMat3 who transforms the vec3 with a mat3.
The parameter type of the matrix in the documentation are false.
It's require a mat3 and in the doc it's a mat4 

Parameters:
Name	Type	Description
out	        vec3	         the receiving vector
a	        vec3	         the vector to transform
m	        mat4         the 3x3 matrix to transform with

Thanks you

toji:
Fixed in #249

PR #249, opened 2017-07-07
Update to use ES2015 syntax and modules

toji:
Pushing this as a pull request in case anyone wants to comment on it before I merge.

The modules conversion is based _very_ heavily on the work done by @bjornharrtell in #209, so most credit goes his way! For the record I initially tried to do the conversion by using ES2015 classes with static methods. (So `class vec3 { static create() { ... } }`) but it turns out that the `<vec>.length()` functions screwed it up because the class types have a default, read only `length` attribute anyway. As a result I reverted back to Bjorn's original method of having each module be a collection of loose functions imported with `import * as <type> from "path"`. I would have preferred the cleaner `import <type> from "path"` that the classes would have given us, but oh well. Maybe in a glMatrix v. 3.0 I'll rename the `length` function to get around it. ;)

Also took the opportunity to manually rebase pull requests #248, #247, #233 into this one to prevent forcing the authors to do that. Should also address #237, #240, #241 and does as much to mitigate #244 as I think this library is likely to do.

Apologies for the slowness of updates! I know there's a lot of other issues still open, and I'll try to get to them as I can, but this is a pretty tough project for me to sink a lot of time into these days. Between work and family I don't have nearly as much time as I used to to maintain personal projects like this one.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "gl-matrix",
   "description": "Javascript Matrix and Vector library for High Performance WebGL apps",
-  "version": "2.3.2",
+  "version": "2.4.0",
   "main": "src/gl-matrix.js",
```
jsantell: I like the pattern of the "main" entry being the `dist/` built file -- this lets others import via node or webpack/browserify still, but not need to assume any sort of ES support (and even if they use babel, they'd need to make sure that they have the same babel configs as you, although limited to the es2015 preset, less of an issue).

An importer would still need the same build environment as the project however if someone is only importing per module (`gl-matrix/src/gl-matrix/mat3` vs `gl-matrix`), but that's the cost to them for selectively importing.

Any reason against using `"main": "dist/gl-matrix.js",`?

toji:
I don't see why not. Thanks for pointing this out! I'll make the change in the next day or so.

bjornharrtell:
I agree with `"main": "dist/gl-matrix.js"` but in addition the `package.json` could also have `"module": "src/gl-matrix.js"`. This will allow tools that understand ES6 modules to use them. See https://github.com/rollup/rollup/wiki/pkg.module.

mreinstein:
@toji  thanks for the update, much appreciated! Looks like some really great changes. A little sad to see browserify go. But webpack is still an improvement. 👍 


> this is a pretty tough project for me to sink a lot of time into these days. 
> Between work and family I don't have nearly as much time as I used to to
> maintain personal projects like this one.

I'm wondering if the maintenance burden could be significantly eased for this project by pulling in all of the modules hosted on stackgl, which split these units out into individual modules? For example, https://www.npmjs.com/package/gl-vec3

Apologies if this has already come up and been decided against or is infeasible. Lately I've been lamenting that as great as the npm ecosystem is we have 164 different representatations for simple math contructs like vectors and matrices.  https://github.com/stackgl/gl-mat3/issues/1

bjornharrtell:
@mreinstein I think split/miniature packages are the unfortunate consequence of tools not able to pull in parts of a package which was indeed not possible before ES6 modules and that it actually increases maintenance burden. https://github.com/rollup/rollup showed that with proper use of ES6 modules there is no longer a need to split a library into multiple modules. AFAIK, webpack also supports this as of version 3 (see https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b).

mreinstein:
> I think split/miniature packages are the unfortunate consequence of tools not able to 
> pull in parts of a package which was indeed not possible before ES6 modules

I wouldn't say it's unfortunate. :) But I'd like to avoid starting a flame war over tiny modules vs larger utility modules. Neither camp is right or wrong, it's just opinion. If you are in the camp of people that like small independent units that do less things, you could use the stackgl variants. If you want more of a fully contained vector/math grab bag, `gl-matrix` could potentially pull in all of those separate modules into 1 index file and make them available via es6. Both camps could get what they want, and it might reduce the maintenance burden on @toji because `gl-matrix` would hopefully just become a simple aggregator for all of those smaller modules.

toji:
Updated "main" and "module" as suggested, thanks!

> A little sad to see browserify go.

Technically browserify was never here. :) I tried it out because I had used a browserify-based workflow in some of my other code that uses modules, but there were some quirks about how it handled this project that I was less enthusiastic about. Fortunately webpack still does the trick with a dose of babel on the side.

> I've been lamenting that as great as the npm ecosystem is we have 164 different representatations for simple math contructs like vectors and matrices.

I don't think I can do anything to actively help that. :) There will never be the "one true math lib" even if vector and matrix math become part of the JS core. Everyone has different needs. That's why I made glMatrix in the first place, because at the time nothing else suited my needs. Selfishly, that's why I'm making this update now, because I want to start using modules more consistently in my other code. The fact that y'all find this library useful too is just a bonus. 😉

mreinstein:
> There will never be the "one true math lib" even if vector and matrix math 
> become part of the JS core. Everyone has different needs.

@toji I think maybe there's some confusion; I'm not advocating for one true math lib. What I'm lamenting is that even within the subset of people that thing `gl-matrix` is really great, There are multiple variants of just that project which have been forked, either because SIMD is unwanted, or because people want more specific modules with smaller scope. https://github.com/stackgl/gl-mat2, https://github.com/stackgl/gl-mat3, https://github.com/stackgl/gl-mat4, etc. These modules are being maintained by brilliant, experienced people (e.g., @mikolalysenko @hughsk and others) Right now those modules have to manually track the changes to gl-matrix to keep parity. You've expressed dissatisfaction with not having time to maintain this. I'm asking if we can come together and maintain this jointly. :) 

lodash is a good example: it is a utility belt with a bunch of useful functions in it. If you want the full grab bag, you can just `import lodash`. Or if you want just debounce, you can `npm install lodash.debounce` and use that in your project without the other stuff. 

Perhaps it's naive, but could `gl-matrix` just be a very simple one file module that `import`s all the independent modules?

PR #254, opened 2017-08-01
fix mat4.targetTo bug

06wj:
The x axis needs to be normalized.


kevzettler:
tests? or bug repro case

toji:
Thanks for the fix, but I agree that we should add a test case when bugs like this are found to ensure we don't regress.

06wj:
@kevzettler @toji  the result matrix.getScaling should be [1, 1, 1], I add some test for it.

Issue #256, opened 2017-08-09
Invocation of not defined function in lookAt


artvinn:
Hey guys, can someone explain me how this line is supposed to work, as it throws me an error? There is no any mat4 object you can refer to. I suppose it should be just `return identity(out)` ?

https://github.com/toji/gl-matrix/blob/a8540ceb6604acf13848d8e796222b02851d8431/src/gl-matrix/mat4.js#L1370

artvinn:
https://github.com/toji/gl-matrix/pull/257

zero298:
I think you're right.  It was probably a miss during 1965b52a3c3f5c05b846245efc9bcbe510924a74 when the `mat4` class was removed.

stefnotch:
@artvinn You can close this issue now.

PR #257, opened 2017-08-09
fix lookAt, mat4 is not defined. #256

artvinn:


amelierosser:
@toji Can you merge this in?

andrevenancio:
pulling this from npmjs still pulls from old version. mat4.js still has vec4.identity.

So if you create a lookAt matrix with both position and target with [0, 0, 0] you bump into this error.. @toji

andrevenancio:
A bit late to the game, but would be nice to add a unit testing for this exception

Issue #260, opened 2017-08-15
Tests broken, pass regardless of what's in the code


jvorob:
It seems like the tests are all passing regardless of what is done to the code.
I tested this by replacing all instances of 
`out[0] = ...`
with
`out[0] = 1;`
in vec2.js, and all the tests still pass, which does not seem to be the intended behavior.

This seems to be an issue as of commit 1965b5 (switching back to webpack)

stefnotch:
Can also reproduce this issue.

kevzettler:
ok confirming this is a real issue

I have isolated an broken a test case:
```
    describe.only("fromValues", function() {
        beforeEach(function() { result = vec2.fromValues(1, 2); });
        it("should return a 2 element array initialized to the values passed", function() { expect(result).toBeEqualish([1, 3]); });
    });
```

produces false positive output
```
> gl-matrix@2.4.0 test /Users/kev/code/gl-matrix
> mocha --compilers js:babel-register --recursive spec



  vec2
    fromValues
      ✓ should return a 2 element array initialized to the values passed


  1 passing (162ms)
```

stefnotch:
@toji This issue is most likely caused by switching to **mocha** from **jasmine**.
https://github.com/toji/gl-matrix/blob/master/package.json 
Commit: https://github.com/toji/gl-matrix/commit/1965b52a3c3f5c05b846245efc9bcbe510924a74#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

Mocha wants the [tests to actually throw an exception](https://mochajs.org/#assertions), in order for them to fail. So, this could be fixed by rewriting spec-helper.js.

kevzettler:
@stefnotch is correct. the issue stems from `spec-helper.js` and I believe the transition from jasmine to mocha. Mocha expects an exception to be raised when asserting.  I have made an attempt to wrap all the expect functions in an assert in https://github.com/toji/gl-matrix/pull/263

stefnotch:
@toji @sinisterchipmunk This issue has been resolved and can be closed

stefnotch:
@jvorob Would you mind closing this issue since it has been resolved.

PR #263, opened 2017-10-18
fix test runner

kevzettler:
in regards to #260 

This wraps the spec-helper.js expect functions in assert calls. This is kind of a clunky fix. An ideal solution would be to migrate to a mocha compatible assertion library. An intermediary solution would be to replace the current expects return values with asserts. I may try that next.

Some of the tests are now failing. This may be a issue with this PR yet or the test may actual failing test cases because the tests were throwing false positives for so long. Will investigate further

### DIFF
--- a/spec/helpers/spec-helper.js
+++ b/spec/helpers/spec-helper.js
@@ -3,65 +3,60 @@ const assert = require('assert');
 
 
 global.expect = function(e) {
-  const asserts = {
-    toBe: function(a) {
-      return e === a;
-    },
-    toEqual: function(a) {
-      return e === a;
-    },
-    toBeDefined: function() {
-      return typeof w !== undefined;
-    },
-    toBeTruthy: function() {
-      return e;
-    },
-    toBeFalsy: function() {
-      return !e;
-    },
-    toBeNull: function() {
-      return e === null;
-    },
-    toBeCloseTo: function(a, p) {
-      return Math.abs(e - a) < (Math.pow(10, -p) / 2);
-    },
-    not: {
-      toBe: function(a) {
-        return e !== a;
-      },
-    },
-    toBeGreaterThan: function(a) {
-      return e > a;
-    },
-    toBeLessThan: function(a) {
-      return e < a;
-    },
-    /*
-       Returns true if `actual` has the same length as `expected`, and
-       if each element of both arrays is within 0.000001 of each other.
-       This is a way to check for "equal enough" conditions, as a way
-       of working around floating point imprecision.
-     */
-    toBeEqualish: function(expected) {
+    return {
+        toBe: function(a) {
+            assert.strictEqual(e, a);
+        },
+        toEqual: function(a) {
+            assert.strictEqual(e,a);
+        },
+        toBeDefined: function() {
+            assert.notStrictEqual(w, undefined);
+        },
+        toBeTruthy: function() {
+            assert(e);
+        },
+        toBeFalsy: function() {
+            assert(!e);
+        },
+        toBeNull: function() {
+            assert.strictEqual(e, null);
+        },
+        toBeCloseTo: function(a, p) {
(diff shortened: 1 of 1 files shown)


kevzettler:
Ok I added a new commit that just changes the spec_helper.js return values into asserts. The test suite passes now with 1 pending. 

Additionally, I verified I could intentionally break a test. Will need another to review and confirm

kevzettler:
@toji could use your eyes and a merge on this. Fairly critical issue.

stefnotch:
A better version would be something that throws the expected value and the actual value. Something like this:

```
toBeFalsy: function() {
  customAssert(!e, false);
}

function customAssert(value, expected) {
  throw `expected ${value} to be ${expected}`;
}
```

stefnotch:
I'm not sure if the current assert messages are always useful. For example, if the lengths mismatch `return assert.fail(e.length, expected.length);` throws a super useful message which tells you something like this: `3 is not 4`

A better version would be something that throws the expected value and the actual value. (That's what jasmine did.)

stefnotch:
@kevzettler Should I create a pull request with a improved version of your spec-helper fixes?

kevzettler:
@stefnotch no need to open a new PR you can push commits directly to the PR branch. 
I think this guide will get you set up for that https://help.github.com/articles/checking-out-pull-requests-locally/

Let me know if you have any problem setting that up. 

Additionally, I can make changes if you leave explicit inline comments so I know what to tweak.
My current understanding is that you feel the error message on:

https://github.com/toji/gl-matrix/pull/263/files#diff-987590674a62c3c54d7ddad3b4837855R51

could be enhanced. Is that correct? Please leave any other comments on inline on the code.

stefnotch:
Code: `expect(true).toBe(false);`

The error message looks like this:
```
      AssertionError [ERR_ASSERTION]: true === false
      + expected - actual

      -true
      +false

```

I'm not really happy with the output. I guess we are either going to have to use a proper library or create our own error messages. Which option do you think is better?

kevzettler:
@stefnotch the issue here is that the `assert` calls are nested within all the `spec_helper.js` `toBe` methods. Traditionally you would use the `assert` calls in the top level test cases and then, per each test case you would pass in the 3rd parameter of `message` which lets you better explain the failed assertion here.

To recreate that experience with this suite. We could add additional `message` parameter to all the `toBe` signatures which then gets passed to the inner `asserts`. The test cases themselves would then need to be individually updated to pass the relevant `message`

Unfortunately, this is slightly more work than removing `spec_helper.js` and just individually instrumenting the test cases with a proper assertion library.

Either way, I feel both options are out of the scope of this PR as the primary goal here is to fix the false positive test suite so that other PRs and updates can continue to happen to the repo. I do feel that improving the assertion messages is a valuable contribution to the repo and would be great for a future PR once this lands.

kevzettler:
Additionally, I have messaged @toji off GitHub for a review. I guess I will start pinging @sinisterchipmunk as well.

stefnotch:
@kevzettler 

After experimenting a bit, it looks like it is possible to adjust the output to be more useful and better. 
Example code:
```
expect(2).toBeCloseTo(20);
```
```
toBeCloseTo: function(a) {
            if(!(Math.abs(e - a) < (Math.pow(10, -EPSILON) / 2))) {
                assert.fail(e, a, `expected ${e} to be close to ${a}`);
            }
        }
```

Error message:
```
      AssertionError [ERR_ASSERTION]: expected 2 to be close to 20
      + expected - actual

      -2
      +20
```


Or, we can throw errors instead of using assert, which gives us complete control over the output.

toji:
Apologies for the delayed response. I understand the concern about getting useful error messages and agree, but it's probably more important to ensure the tests are actually failing properly first. As such I'm merging this PR now and further improvements to the messages can be done in a follow up.

kevzettler:
closes #260

PR #264, opened 2017-10-22
Link to a tutorial because official site has no examples of use

jameshfisher:


toji:
I should really finish up the tutorial page I was working on, but I don't object to having this link in place in the meantime. Thanks.

PR #271, opened 2017-11-11
Adding outer product to matrix modules.

ptpham:


stefnotch:
My mathematics background isn't too strong, so, would you mind explaining what an outer product is? And, more importantly, when do you need it?

ptpham:
Note that the outer product I'm referring to is an algebraic reformulation of the product of P transpose and Q. Instead of doing that matrix multiplication, it is equivalent to do a sum of outer products of the rows of P and Q.

stefnotch:
@ptpham That's interesting. :) However, I'm not sure how many people need this feature.

What do you think about creating an issue where all sorts of features can be requested and voted on? And, if enough people are interested in something, it may get added to the library.

ptpham:
Haha. Well, it's okay. I made a fork a while ago anyways to better suit my use cases. XD Thanks for the interest. =P

Issue #272, opened 2017-11-14
is repo being maintained ?


andrevenancio:
I've been following a couple of `Issues` and `Pull requests` that interest me. And its been a while since I see any activity in the repo, from @toji despite the `tags`. I for example, have a 2 week old PR with no comment from any users / admins.

I'm making myself available to help in any way possible. either to test stuff, or to help setting up priorities, but we're getting to a point where a lot of things are pilling up.

I'm already "living" from a fork from this repo with some change to fix a couple of issues, but I don't think this is the right approach.

Again, happy to help where I can.

stefnotch:
It might be a good idea to create and maintain your own fork, since @toji seems to be quite busy most of the time. Also, this repository has a huge backlog of issues which were never closed, even though they should be.

@toji What do you think?

andrevenancio:
I really think that's not the best way... People don't need to be looking for a branch from this repo and see which one is better maintained. I understand the work load might be a lot, but its up to @toji and @sinisterchipmunk to either allow collaborators who can spend more time here and close issues, review PRs merge changes, or just make a yearly review or something 😃

andrevenancio:
Yeah I know, I remember being on a chat with you on another issue 😛 There's definitely people into gl-matrix 😛

andrevenancio:
Now I've noticed that someone added a PR that does the same as an old PR I've submitted. 
Which is exactly the point of this issue. We need some love on this repo...

PR #268, opened 2017-10-31
Deprecation Warnings

andrevenancio:
Fixing deprecation warning when using as npm module.

https://github.com/toji/gl-matrix/issues/267

Issue #267, opened 2017-10-31
Deprecation Warning when using as npm module


andrevenancio:
the dependency `babel-preset-es2015` as been deprecated, more info [here](http://babeljs.io/env).

Anyone importing `gl-matrix` via npm modules, will be importing the src as specified on package.json "module" property ("module": src/gl-matrix.js)

This causes a warning in the terminal console. not breaking change, but creates some annoyance to some users.

Should be an easy fix to substitute the `babel-preset-es2015` to the new `babel-preset-env`.

Creating a PR shortly

andrevenancio:
PR at https://github.com/toji/gl-matrix/pull/268

andrevenancio:
Did anyone had a chance to review this?

PR #275, opened 2017-12-04
Replace babel-preset-2015 with babel-preset-env

TimvanScherpenzeel:
Babel has moved to a system of instead of using yearly presets to a more long term solution that incorporates all yearly presets in a single preset. This helps with creating a more future proof solution.

This pull request replaces the current `babel-preset-es2015` with `babel-preset-env` for the reasons as explained in the dialog upon install:

> We're super 😸  excited that you're trying to use ES2015 syntax, but instead of making more yearly presets 😭 , Babel now has a better preset that we recommend you use instead: npm install babel-preset-env --save-dev. preset-env without options will compile ES2015+ down to ES5 just like using all the presets together and thus is more future proof. It also allows you to target specific browsers so that Babel can do less work and you can ship native ES2015+ to user 😎 ! We are also in the process of releasing v7, so please give http://babeljs.io/blog/2017/09/12/planning-for-7.0 a read and help test it out in beta! Thanks so much for using Babel 🙏, please give us a follow on Twitter @babeljs for news on Babel, join slack.babeljs.io for discussion/development and help support the project at opencollective.com/babel

This has the advantage of packages that build parts of `gl-matrix` not having to install `babel-preset-es2015` just for `gl-matrix` when the rest of the build tools are already using `babel-preset-env`.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015"]
+  "presets": ["env"]
 }
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "devDependencies": {
     "babel-core": "^6.25.0",
     "babel-loader": "^7.1.1",
-    "babel-preset-es2015": "^6.24.1",
+    "babel-preset-env": "^1.6.1",
     "babel-register": "^6.24.1",
     "jsdoc": "^3.4.3",
     "mocha": "^3.4.2",


andrevenancio:
duplicated of my [PR](https://github.com/toji/gl-matrix/pull/268) see also [this](https://github.com/toji/gl-matrix/issues/267)

Though nobody seems to [car](https://github.com/toji/gl-matrix/issues/272) about updates :(

PR #281, opened 2018-02-07
add .js extensions to imports to allow vanilla loading

damienmortini:
Omitting `.js` extension in imports prevent gl-matrix to be used natively with `<script type="module">`.
Figured that it doesn't harm to add them ;)


shannon:
@damienmortini Was just going to submit the same PR. The updated webpack UMD is important too because it gets around this issue: https://github.com/webpack/webpack/issues/5843.

You may want to remove the package-lock.json file from this though as that's not really encouraged for libraries. And just update the package.json so it has `"webpack": "^3.10.0"`. This was the version that fixed that issue.

Just my two cents.

damienmortini:
@shannon Thanks, very useful, I will check the new webpack.
Also good call for the package-lock, it makes sense!

yanchith:
On a related note: Is anyone interested in having es6 module bundle (instead of webpack-umd) for gl-matrix? I am creating one in my fork atm, so if there is interest, I would be willing to contribute.

PR #293, opened 2018-04-24
Skew mat2d

talltyler:
Add a simple function to skew a mat2d.


stefnotch:
Unit tests would be nice

PR #296, opened 2018-04-28
Should move `npm` to devDependencies

zixia:
instead to put it into `dependencies` @stefnotch 

I'm upgrading Node to v10 and NPM to v6 yesterday but I found that this module installs a local out-dated npm in my `node_modules` directory which does not compatible with Node v10.

I do not understand why we need to install an npm module localy, but I believe that move it to `devDependeicnes` will solve my problem(and potential others')

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "babel-register": "^6.24.1",
     "jsdoc": "^3.4.3",
     "mocha": "^4.0.1",
+    "npm": "^5.8.0",
     "node-libs-browser": "^2.0.0",
     "webpack": "^3.11.0"
   },
@@ -42,6 +43,5 @@
     "build-all": "webpack && webpack --config webpack.config.min.js"
   },
   "dependencies": {
-    "npm": "^5.8.0"
   }
 }


stefnotch:
I'm not very well versed with `npm` and thus, I'll wait for some other opinions regarding this.

nkbt:
Neither of files in `node_modules/gl-matrix/src` use `npm` so it definitely should not be in `dependencies`.

I am not sure `npm` should be even `devDependency`, maybe `optional` or `peer`. Because it blows up `lock` files for pretty much no reason.

zixia:
@nkbt Yes, I agree with you that it feels better if we could remove the NPM from all the dependencies.

So could we merge this PR now? or I have to write a helper/hacky script for my build process under Node v10 ...

nkbt:
I'd say this one is a 100% safe change, then we can do some better one with peer/optional as a next step (as well as `.npmignore` to remove all the test/dev/ stuff from dist package)

nkbt:
On a side note, why this is important: we track licenses of all deps from production dependencies, so keeping `npm` in `dependencies` adds a gazillion of new extra licenses (with a lot of broken ones too)

stefnotch:
Ok, merged. 👍

PR #307, opened 2018-05-16
Move NPM to `peerDependencies`.

### LINKED ISSUE #300: Move npm to optional
Currently, `npm` is a devDependency. However, it could be moved to [optionalDependencies](https://github.com/toji/gl-matrix/pull/296#issuecomment-386962650). 
https://github.com/toji/gl-matrix/blob/d9eb397c14b93581247916c2f807fc54f1cb597e/package.json#L32

daaain:
Fixes #300.

I'm wondering though if it's needed to be listed as a dependency at all? I did a repo wide search and it seems the only dependency on NPM is being able to do `npm run`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,11 @@
     "babel-register": "^6.24.1",
     "jsdoc": "^3.4.3",
     "mocha": "^4.0.1",
-    "npm": "^5.8.0",
     "node-libs-browser": "^2.0.0",
     "webpack": "^3.11.0"
+  },
+  "peerDependencies": {
+    "npm": "^5.8.0"
   },
   "license": "MIT",
   "scripts": {


daaain:
Thanks a lot @stefnotch for the quick turnaround! Would you mind bumping the package version and pushing a release to NPM?

(As an aside I still think it would have been better to just get rid of it altogether as a dependency...)

stefnotch:
@daaain Sure!

(I'm not sure why it's a dependency. toji might know the reason.)

daaain:
The docs on that link is about enforce a Node version, not NPM, as far as I understand. For that a more standard approach would be adding the Node version to `engines` in `package.json` which is also mentioned in there.

stefnotch:
🤦‍♂️ I should have properly read through it first.

Issue #311, opened 2018-05-18
 ReferenceError: window is not defined


ouyangzhigang:
nodejs-dashboard -- node -r nodejs-dashboard web-test-server.js
E:\nodejs\npm-global\node_modules\nodejs-dashboard\node_modules\_gl-matrix@2.6.0@gl-matrix\dist\gl-matrix.js:38
})(window, function() {
   ^

ReferenceError: window is not defined
    at Object.<anonymous> (E:\nodejs\npm-global\node_modules\nodejs-dashboard\node_modules\_gl-matrix@2.6.0@gl-matrix\dist\gl-matrix.js:38:4)
    at Module._compile (module.js:577:32)
    at Object.Module._extensions..js (module.js:586:10)
    at Module.load (module.js:494:32)
    at tryModuleLoad (module.js:453:12)
    at Function.Module._load (module.js:445:3)
    at Module.require (module.js:504:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (E:\nodejs\npm-global\node_modules\nodejs-dashboard\node_modules\_drawille-canvas-blessed-contrib@0.1.3@drawille-canvas-blessed-contrib\index.js:3:16)
    at Module._compile (module.js:577:32)

stefnotch:
https://github.com/toji/gl-matrix/issues/310
^Same issue

@andrevenancio Looks like the PR did cause some issues?

stefnotch:
https://github.com/toji/gl-matrix/blob/e5b9a08ee61967dfa852d69f2d96d3c606767e64/dist/gl-matrix.js#L38

`window` used to be `this`

andrevenancio:
Hey @stefnotch as I comment on #310 the module has been there since version 2.4.0.

However I've noticed there's a [bug](https://github.com/webpack/webpack/issues/6522) being reported in webpack github. Bare with me, I'll sort out a quick fix until they fix their bug

stefnotch:
This is hopefully fixed now.

Issue #310, opened 2018-05-18
package.json has broken "main" field


Fuzzyma:
When you `npm install` the package as is and use `require('gl-matrix')`, this library will not work in nodejs anymore because at some point the `main` field in the package.json was changed to `dist/gl-matrix` where it should be `src/gl-matrix`.
You dont want to use the webpacked version when you use node.
Please correct this because otherwise nobody can use this package anymore

stefnotch:
(Relevant issue: https://github.com/toji/gl-matrix/issues/277 )

@andrevenancio You probably know more about this than I do.

andrevenancio:
Hi @Fuzzyma, you are right, the `main` property in the `package.json` points to the "compiled" version. I don't know why `module` was added. It was introduced by @toji with this [commit](https://github.com/toji/gl-matrix/commit/7850f1f1e09be244f3bf84aa7208e45259b4d9b1)

From what I can ascertain the `module` property is meant to be used by webpack and rollup as ES6 modules (allowing three shaking).

With that said, I just want to point out that the exported version on the `dist/` folder (`dist/gl-matrix.js`) though it has some comments introduced by webpack, its not a `webpack` version. Webpack is exporting with the UMD option which exposes the library under all the module definitions, allowing it to work with CommonJS, AMD and as global variable.

We can remove the comments from that version (and we should @stefnotch ) but still the only thing that's happening is that you're merging all the files into one.

Unless I'm missing something this should not be a problem.

andrevenancio:
@Fuzzyma  Can you pull my version https://github.com/toji/gl-matrix/pull/312 and give it a go on your node app?

I've tested this locally with something very simple, and it seems to work

test.js
```
var vec3 = require('gl-matrix').vec3;
console.log(vec3.create());
```

and running
```
node test.js
```

returns the expected `Float32Array [ 0, 0, 0 ]`

Give it a go too @stefnotch

Fuzzyma:
the `windows` was were my code was breakig. However this was just reported to my repo. I cannot test it atm but I am pretty sure it should work now. I will report back when my people close the issue on my repo

PR #312, opened 2018-05-18
Fixing webpack4 bug introducing `window` object to UMD modules

andrevenancio:
as pointed out by https://github.com/toji/gl-matrix/issues/310 and https://github.com/toji/gl-matrix/issues/311

Importing the latest version to be use in a node.js application breaks. This is because of a reported bug on webpack4 https://github.com/webpack/webpack/issues/6522


stefnotch:
Thank you!

PR #315, opened 2018-05-28
Fix expect.toBeEqualish and vec4.random bug

06wj:
```toBeEqualish``` not check the ```NaN``` value.
```vec4.random``` return `NaN` value, it`s wrong but pass the test.


stefnotch:
Thank you!

PR #318, opened 2018-06-09
Add infinite perspective matrix

d07RiV:
Support for zFar = infinity for projection matrix.

Dunno why dist build ended up so different.


stefnotch:
@d07RiV Wouldn't it be cleaner to simply modify https://github.com/toji/gl-matrix/blob/f1871a7c4495d441573258bfdf995774da08797f/src/gl-matrix/mat4.js#L1268 to support setting the far value to `Math.Infinity`?

d07RiV:
That seems less intuitive to me, but if you think that's better then maybe.

stefnotch:
@d07RiV Well, I'd prefer to not add a ton of functions that might not be used as often.
`less intuitive`: Well, other than adding another function, how could you make it more intuitive?

d07RiV:
You think null value for `far` is better?

stefnotch:
@d07RiV That seems equally good/bad. I'd say, let's just go with Math.Infinity and add a note in the documentation. Something like `@param {number} far Far bound of the frustum, supports Math.Infinity`

src/gl-matrix/mat4.js
```diff
@@ -1278,42 +1278,18 @@ export function perspective(out, fovy, aspect, near, far) {
...
-  out[7] = 0;
-  out[8] = 0;
-  out[9] = 0;
-  out[10] = -1;
-  out[11] = -1;
-  out[12] = 0;
-  out[13] = 0;
-  out[14] = -2 * near;
   out[15] = 0;
+  if (far != null && far !== Infinity) {
```
stefnotch: That's also fine. 🙂 If anyone has an issue with it, they can just open an issue.

d07RiV:
Why not both?

d07RiV:
Also I'm not sure why build was changed so much, and I didn't generate docs because it would put my time/timezone in every file.

stefnotch:
@d07RiV Well, the checks passed and the PR adds some useful functionality, so, merging this!
Thank you! 👍

PR #335, opened 2018-08-14
Provide ES modules with other ES features transpiled to ES5.

### LINKED ISSUE #306: CRA/UglifyJS fails to minify common.js
When using the module in a CRA app, or any app that uses UglifyJS as a build step, I get the following message:

    Failed to minify the code from this file:

        ./node_modules/gl-matrix/src/gl-matrix/common.js:28

And links me to this:

https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-build-fails-to-minify

Which, in turn, suggests that we file an issue and ask that the package be published pre-compiled.

dtinth:
Fix #306

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,13 @@
 {
-  "presets": ["env"]
+  "env": {
+    "development": {
+      "presets": ["env"]
+    },
+    "production": {
+      "presets": ["env"]
+    },
+    "esm": {
+      "presets": [["env", { "modules": false }]]
+    }
+  }
 }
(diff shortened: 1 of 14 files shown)


dtinth:
I have tested this with a `create-react-app` app and confirmed that it works, including tree shaking. In other words, only the functions that are actually used are included in the final bundle:

image

If you want to try, use this in `package.json`:

```
    "gl-matrix": "dtinth/gl-matrix",
```

stefnotch:
Looks good to me.

I'm going to wait a bit before pushing it to npm, just in case any issues arise.

stefnotch:
Ok, published it to npm.

PR #339, opened 2018-11-01
Allow cherry picking both cjs/esm

TrySound:
Importing from entry point is useful when everything is module is just
functions. However this packages reexports namespaces which blocks
treeshaking of all functions.

In this PR I suggest to change entry points to cjs and esm and provide
same file structure as in source.

This will allow in project like [math.gl](https://github.com/uber-web/math.gl) swap cjs with esm on build step without importing evrything from `gl-matrix`.


TrySound:
Something is broken in webpack build

stefnotch:
@TrySound Ok, I fixed the issue with the Webpack build. 

Regarding your PR, will it potentially break various libraries and projects that depend on gl-matrix? (If yes, I'm fine with doing a major version bump.)

TrySound:
Yeah, I think it may be a breaking change in some cases

TrySound:
I would also a bit flatten file structure

stefnotch:
Dear @andrevenancio , since you know a fair bit more about Webpack and whatnot than I do, would you mind taking a look at this PR?

andrevenancio:
Just saw this now. Will have a look shortly.

andrevenancio:
> Something is broken in webpack build
What do you mean @TrySound ? Any more info on that?

Having a look at gl-matrix again, made me realise some stuff about the library:
a) we're exporting the `gl-matrix.js` and the `gl-matrix.min.js` exactly with the same content. I think ever since Webpack4 introduced the `mode` property, they now minify everything marked with `mode: "production"` so our `gl-matrix.js` file is actually minified. I can create a PR that fixes this by overriding the UglifyJS default settings to make it readable. @stefnotch ?

b) We probably need to add an example on the README.md file for importing this. maybe an ES5 code example and an ES6 code example. 

In regards to this PR itself, I'm either looking at the wrong place, or something seems to be missing.
On the `package.json` you have `"main": "cjs/index.js"` but you dont have a folder called `cjs` on your build. Am I missing something?

TrySound:
@andrevenancio 

a) It's usually simply splitted with development and production bundles without overriding uglify options.

b) There are both esm and cjs. But they should be created only before the package is published. Build artefacts are usually not committed to git.
See here https://github.com/istarkov/rifm and here https://unpkg.com/rifm@0.4.3/

andrevenancio:
Of course I am missing something... `npm run build-all` lol :) Yeah seems to build all properly, tests also passing. I was under the impression that only rollup uses the property "module" on the package.json which means anyone using webpack will effectively be importing the `cjs/index.js` version right?

andrevenancio:
> @andrevenancio
> 
> a) It's usually simply splitted with development and production bundles without overriding uglify options.
> 
> b) There are both esm and cjs. But they should be created only before the package is published. Build artefacts are usually not committed to git.
> See here https://github.com/istarkov/rifm and here https://unpkg.com/rifm@0.4.3/

You reckon we should build it with `mode: development` is that what you're suggesting?

TrySound:
b) Currently switching can be achieved on user side by some code transformers
```
import * as mat2 from 'gl-matrix/cjs/mat2';
// transform to
import * as mat2 from 'gl-matrix/esm/mat2';
```

But I have an idea how to route between esm and cjs. I'm gonna push it in the next PR after merging this one.

TrySound:
@stefnotch Yes, `development` and `production` modes.

Actually not only rollup. webpack, rollup and parcel are all handles `module` field.
https://webpack.js.org/configuration/resolve/#resolve-mainfields

TrySound:
It's a kind of standard in user land.

andrevenancio:
> ```
> import * as mat2 from 'gl-matrix/cjs/mat2';
> import * as mat2 from 'gl-matrix/esm/mat2';
> ```

Got it!.

Yeah, this will introduce breaking changes to users using older versions.
Which is not the end of the world, but probably needs to be documented on the README.md.

I would thought that when we're importing gl-matrix (because we're pointing to the src folder) the tree shaking would happen. Is this not the case @TrySound ?

TrySound:
Treeshaking with importing from `gl-matrix` which points to `gl-matrix/dist/esm/index.js` is not possible because reexporting namespaces as objects converts them to this
```js
export var math2 = Object.freeze({
  add: () => {}
})
```

That's why path imports is the only good way here. And we may achieve it with some the trick I'm gonna implement in the next PR.

```
gl-matrix/mat2/package.json
{
  "gl-matrix/mat2",
  "main": "../dist/cjs/mat2.js",
  "module": "../dist/esm/mat2.js"
}
```
With this file importing from `gl-matrix/mat2` will route bundles to esm and node will be routed to cjs.

andrevenancio:
> Treeshaking with importing from `gl-matrix` which points to `gl-matrix/dist/esm/index.js` is not possible because reexporting namespaces as objects converts them to this
> 

I was asking about the current version we have. not your PR in particular. The current `gl-matrix` version on npm points to `lib/gl-matrix.js` shouldnt it point to `src/gl-matrix`?

I'm getting confused why we have `/lib` and `/dist` 

Wouldnt we be better off with

`src/` (ES6)
`dist/` (ES5) 
`lib/` (removed)

?

TrySound:
Currently lib is esm files, dist is a pair of UMD bundles. They are useful too for in browser editors.
This is my holy graal of distribution https://unpkg.com/rifm@0.4.3/dist/

My suggestion in this PR is esm+cjs+dist.

Pointing to sources is not yet actively used. Maybe with "esnext" field.

The point is that `esm` is not `es6`. `esm` is still transpiled but without cjs transformation to allow users to skip they babel setup for node_modules. `esm` can be transpiled in better way by modern bundlers.

Well yeah, `export * from './vec2'` would simplify everything and I would propose only bundles like in rifm example. However this will require to change function names to prefixed.
https://github.com/toji/gl-matrix/blob/master/src/gl-matrix/mat2.js#L13
Are you ready to do this?

TrySound:
Rollup is better for umd in this case. I may migrate from webpack in the next PR.

TrySound:
So would you like to go with this changes first? Cherry picking will be the best for consumers with current api.

andrevenancio:
I will leave this up to @stefnotch to decide. Not sure if we're helping more people than harming with forcing them to update their libraries.

TrySound:
This will unblock uber projects like react-map-gl from using this project. Currently they are use a fork where each package is vec2, mat3 etc.

Im sure users will be happy to consume less parts of this library and get it with less config.

stefnotch:
@TrySound I'm going to merge this PR and publish it as a pre-release version to NPM in order to not break anything for anyone. (something like v3.0.0 pre release) Does that sound like a good idea?

TrySound:
Yeah, starting with prerelease is a good idea. We have some more work to do.

andrevenancio:
Let me know if I can help.

TrySound:
https://github.com/toji/gl-matrix/pull/343

PR #342, opened 2018-11-05
Travis CI fix

stefnotch:
[This might fix it](https://github.com/plotly/dash-component-boilerplate/issues/12#issuecomment-425040481)


stefnotch:
Seems to fix it. Wohoo!

PR #343, opened 2018-11-11
Add routes to simplify cherry picking

TrySound:
`main` and `module` allows us to provide an entry points for both
bundler and node. However this doesn't work with modules cherry picking.

In this diff I suggest to add `package.json` with `main` and `module`
fields to each module. This will allow to import this library like this

```js
import { create } from 'gl-matrix/vec2';
```

The structure looks like this

```
README.md
LICENSE.md
cjs/*.js
esm/*.js
mat2/package.json
mat2d/package.json
mat3/package.json
mat4/package.json
quat/package.json
quat2/package.json
vec2/package.json
vec3/package.json
vec4/package.json
gl-matrix-min.js
gl-matrix.js
```

Each `package.json` looks like this

```json
{
  "name": "gl-matrix/vec2",
  "main": "../cjs/vec2.js",
  "module": "../esm/vec2.js"
}
```

To prevent root folder pollution I made it private and copy necessary
files like `package.json`, `README.md` and `LICENSE.md` into dist
folder. To publish package you need to go first to `dist` folder.


TrySound:
/cc @stefnotch 

After this one I'm gonna to send one last PR to improve umd bundles.

andrevenancio:
Seems to be taking shape nicely! Nice one!

TrySound:
@stefnotch Can this be merged?

stefnotch:
Oh, sorry for the late response. Yes, this does look quite fine, so, I'm merging it.

PR #350, opened 2019-01-04
Adding Vector.zero method

maierfelix:
Adding method to set all components of a vector to zero.

I often encounter this when working with cached vectors. Having a ``Vector.zero`` method is a more clear and probably faster way than calling ``Vector.set(out, 0, ..)``.


stefnotch:
Great PR!

Why it's great and I'm merging it
✔️  Convenience: resetting cached vectors and creating zeroed out vectors
✔️ Analogous to existing functions: matrix `identity`
✔️ Unit tests: present

 Performance: unknown, possibly faster than the alternative

maierfelix:
Thanks!

Issue #323, opened 2018-06-26
Ship types (index.d.ts) to work with Typescript


jhm-ciberman:
Hello. I want to propose to ship the types.d.ts file to work with typescript with the gl-matrix. 

## Why? 
Because in this way the types could be updated with each change. For example. The version that is now in DefinetelyTyped (The place with all the types definitions for all npm packages) is outdated and not contain the new mat4.targetTo(....) method. 

## How? 
Simply copy this file: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/gl-matrix/index.d.ts

to the gl-matrix repo root, and then add the `types` field to the package,json:
```
{
    "name": "awesome",
    "author": "Vandelay Industries",
    "version": "1.0.0",
    "main": "./lib/main.js",
    "types": "./index.d.ts"
}
```

More instructions here: http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Then npm publish the new version. And that's all.. you will have thousands of Typescript developers happy. :) 

If you want, you can also add the mat4.targetAt(...) method to the index.d.ts. It should be easy to add. Also, you can add another missing methods if you wish.

Epskampie:
Hey @jhm-ciberman, have you considered contributing to the definitelytyped definition by adding the mat4.targetTo definition? This way the developers have more time to just work on gl-matrix.

It's quite easy, just make a pull request in the definitelytyped repo: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/gl-matrix . They usually get merged in a week or so. Here's a guide: https://github.com/DefinitelyTyped/DefinitelyTyped#how-can-i-contribute

dcbrwn:
@Epskampie definitely typed project is deprecated. @jhm-ciberman proposes correct way to do this. Btw gl-matrix devs already kinda work on typings by specifying JSDocs. TS compiler checks them, but since they are a bit incomplete it often complains about type mismatches.

stefnotch:
> but since they are a bit incomplete it often complains about type mismatches

That's a problem that I can, and should definitely address. Any examples of types that are incomplete/wrong?

dcbrwn:
@stefnotch not sure what causes this, but when type is explicitly declared for a variable, typescript built into webstorm complains that it's not correct. if type is not declared and inferred - all is just fine.

![image](https://user-images.githubusercontent.com/12214069/45143747-c153c880-b1c4-11e8-8efa-5c04e5398895.png)

[...]

PR #351, opened 2019-01-05
Update JSDocs to allow correct type checks

dcbrwn:
An attempt to address #323. This PR renames all data types to Pascal case and adds @typedefs describing them. It's the cheapest solution I came up with that requires minimum maintenance.

Also I intentionally renamed types in docs. Is this okay?

Typechecks now working a bit better relying solely on JSDocs:
```typescript
import { vec2, vec3 } from './src/index';

let a: vec2.Vec2;

a = vec3.create();

/* tsc outputs

Type 'Float32Array | [number, number, number]' is not assignable to type '[number, number] | Float32Array'.
  Type '[number, number, number]' is not assignable to type '[number, number] | Float32Array'.
    Type '[number, number, number]' is not assignable to type '[number, number]'.
      Types of property 'length' are incompatible.
        Type '3' is not assignable to type '2'.

*/
```

### DIFF
--- a/docs/mat2.js.html
+++ b/docs/mat2.js.html
@@ -36,7 +36,7 @@
 /**
  * Creates a new identity mat2
  *
- * @returns {mat2} a new 2x2 matrix
+ * @returns {Mat2} a new 2x2 matrix
  */
 export function create() {
   let out = new glMatrix.ARRAY_TYPE(4);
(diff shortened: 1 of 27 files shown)


dcbrwn:
Caveat: tuple syntax requires at least typescript 3.

stefnotch:
This seems to be a pretty decent solution. (I took a look at it and couldn't figure out anything better, or anything that would allow us to use `vec2` as a type **and** a module instead of having to type `vec2.Vec2`)

stefnotch:
I found a way to get type checking to work with TypeScript that requires basically no maintenance. 😄 

```typescript
// b.ts
import * as mat3i from "./mat3.js";
//import { mat3 } from "./mat3.js";

let mat3: typeof mat3i = mat3i;
type mat3 = number[];

export { mat3 };
```

stefnotch:
Did you manage to build the documentation with this?

stefnotch:
I'm currently working on a better solution. It's possible that the best solution would be to make the switch to TypeScript.

Issue #225, opened 2016-09-29
lookAt is backward


greggman:
## tl;dr
- `targetTo` creates a matrix that makes something look at something else
- `lookAt` works like gluLookAt() and creates the inverse of that matrix

## Original Comment
I know this is subjective but lookAt is not generating a matrix that makes something _lookAt_ something else. It's making the inverse of that. A matrix that moves everything else such that the `eye` is at the origin. 

Why is that backward? Because in a typical scene you might have 1 camera + 50 players +100 gun turrets. You want 50 NPC heads to look the the nearest other NPC. You want the 100 turrets to target the closest NPC. That's 150 calls to lookAt except you can't use glMatrix's lookAt function for this unless you call invert on 150 matrices. There's only 1 camera so it makes more sense for `lookAt` function to make a matrix that puts the `eye` at `eye` and orients it to _lookAt_ at `target`. You then have a useful function and you can invert for that one special view matrix case when after you've made the camera _lookAt_ it's target you need view matrix for that camera to render from its point of view.

httpdigest:
This is just how the very common GLU gluLookAt() function is defined, that has been around for more than two decades, and has been copied/adapted in various libraries for various programming languages.
If glMatrix would change its meaning, then all people knowing OpenGL/GLU and all other libraries/languages would feel uncomfortable.
You could however introduce a new function, which just uses the column vectors that lookAt also computes, and then does not transpose/invert them into row vectors but just store them in the output matrix as column vectors. And then set the translation part just to "eye" (position).
This would be quite simple to do.

stefnotch:
How should that alternative function be named? (I would be willing to implement it.)

andrevenancio:
I have to agree with @greggman point. Though I understand the historic reasons pointed out by @httpdigest. So regardless of opinions it would be useful to have this method in the library. `targetTo` can be an alternative. I've written a small example, but it assumes 3 vec3 are defined somewhere on the page. Might actually need to simplify the vec3 calculations and do them all inline like the method lookAt currently is doing. 

```
lookAt(out, eye, target, up) {
        vec3.subtract(zAxis, eye, target);
        vec3.normalize(zAxis, zAxis);
        vec3.cross(xAxis, up, zAxis);
        vec3.cross(yAxis, zAxis, xAxis);

        out[0] = xAxis[0];
        out[1] = xAxis[1];
        out[2] = xAxis[2];
        out[3] = 0;
        out[4] = yAxis[0];
[...]

PR #357, opened 2019-03-21
optim: Math.hypot replace Math.sqrt(Math.pow(x,2))

jack-shanghai:
~~it seems targetTo function have some problems, according to my analysis, we should invert the 'out' matrix.~~ Besides, I replace Math.sqrt(Math.pow()+...) with Math.hypot() which is more elegant.

### DIFF
--- a/src/mat2.js
+++ b/src/mat2.js
@@ -291,7 +291,7 @@ export function str(a) {
  * @returns {Number} Frobenius norm
  */
 export function frob(a) {
-  return(Math.sqrt(Math.pow(a[0], 2) + Math.pow(a[1], 2) + Math.pow(a[2], 2) + Math.pow(a[3], 2)))
+  return(Math.hypot(...a))
 }
 
 /**
--- a/src/mat2.js
+++ b/src/mat2.js
@@ -291,7 +291,7 @@ export function str(a) {
  * @returns {Number} Frobenius norm
  */
 export function frob(a) {
-  return(Math.hypot(...a))
+  return(Math.hypot(a[0],a[1],a[2],a[3]))
 }
 
 /**
--- a/src/mat4.js
+++ b/src/mat4.js
@@ -1536,7 +1536,7 @@ export function str(a) {
  * @returns {Number} Frobenius norm
  */
 export function frob(a) {
-  return(Math.hypot(...a))
+  return(Math.hypot(a[0],a[1],a[3],a[4],a[5],a[6],a[7],a[8],a[9],a[10],a[11],a[12],a[13],a[14],a[15]))
 }
 
 /**
--- a/src/mat4.js
+++ b/src/mat4.js
@@ -1477,8 +1477,9 @@ export function targetTo(out, eye, target, up) {
       z1 = eyey - target[1],
       z2 = eyez - target[2];
 
+  let len = z0*z0 + z1*z1 + z2*z2;
   if (len > 0) {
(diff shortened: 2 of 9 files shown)


src/mat2.js
```diff
@@ -291,7 +291,7 @@ export function str(a) {
  * @returns {Number} Frobenius norm
  */
 export function frob(a) {
-  return(Math.sqrt(Math.pow(a[0], 2) + Math.pow(a[1], 2) + Math.pow(a[2], 2) + Math.pow(a[3], 2)))
+  return(Math.hypot(...a))
```
stefnotch: I would prefer to not use the spread operator, as I'm not entirely sure about it's performance.

jack-shanghai: that ok, how about Math.hypot to replace Math.sqrt(Math.pow)

stefnotch:
Regarding targetTo, it's supposed to be this way. For more information, check out https://github.com/toji/gl-matrix/issues/225

However, the Math.hypot optimization seems to be pretty interesting. 😄

src/mat2.js
stefnotch: Yes, of course. That's pretty cool!

src/mat4.js
```diff
@@ -1513,7 +1512,7 @@ export function targetTo(out, eye, target, up) {
   out[13] = eyey;
   out[14] = eyez;
   out[15] = 1;
-  return out;
+  return invert(out, out);
```
stefnotch: As stated in #225 , this does not need to be inverted. If you want the inverse, you should use `lookAt`

jack-shanghai: ok, due to my poor English, I still can't understand #225 said, it seems lookAt function has achieved the goal which I thought target should would do, Thank you!

stefnotch: @jack-shanghai Yes, that is correct. 

Would you mind changing this pull request from `fix: targetTo..` to `optim: Math.hypot..`?

jack-shanghai: sorry, I forget it, it's done

stefnotch:
👍 Thank you for the great pull request! 😄 I'll merge it ASAP.

doodlewind:
`Math.hypot` is interesting indeed 👍 A concern is that is gl-matrix supposed to be compatible with IE? `hypot ` is landed in ES2015 so it may not work on IE11.

stefnotch:
Yes, that is a valid concern indeed. There are a few options here:
- Drop IE support: Sub-optimal
- Reject the PR: Also sub-optimal, as gl-matrix is supposed to be super fast
- Transpile `Math.hypot` to `Math.pow(Math.sqrt(...`
- Ship a `Math.hypot` polyfill with gl-matrix

doodlewind:
How about detecting the browser's `Math.hypot` support and use our current implementation as fallback? In this way we don't have to include the heavy official polyfill libs, neither do we need extra transpile process.

stefnotch:
> heavy official polyfill libs

I didn't suggest adding a fully blown polyfill library. We only have to polyfill, `Math.hypot`, which is [rather easy and lightweight](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot#Polyfill). 😄 
However, the polyfill option does have a disadvantage: gl-matrix is going to be slower on IE. But I suppose that's an acceptable downside.

jack-shanghai:
How can you be so patient, it's my pleasure to do my bit!

Issue #360, opened 2019-05-10
Seems like the identity for mat2d is set incorrectly?


baseten:
It looks like the identity is being set incorrectly both on https://github.com/toji/gl-matrix/blob/master/src/mat2d.js#L36 and https://github.com/toji/gl-matrix/blob/master/src/mat2d.js#L84 

Correct me if I'm wrong but should it not be `out[3] = 0; out[4] = 1;`? It seems like `b` not `d` is being set to 1

baseten:
Ah, sorry I see now this is a column/row major order thing. Is the doc bloc the culprit? https://github.com/toji/gl-matrix/blob/master/src/mat2d.js#L10

stefnotch:
Yep, the documentation is apparently wrong. In the documentation of that class, the columns and rows have been switched.
https://github.com/toji/gl-matrix/blob/9ea87effc0ca32feb8511a01685ee596c6160fcc/src/mat2d.js#L107

Feel free to create a PR

baseten:
Is this intentional as per note on [homepage](http://glmatrix.net/)? I understand wanting to follow OpenGL convention for column major examples, but since this appears to be a code block it's extra confusing 😄

stefnotch:
The code is intentional, the mat2d documentation comment is wrong. A good example is [the translate function](https://github.com/toji/gl-matrix/blob/9ea87effc0ca32feb8511a01685ee596c6160fcc/src/mat2d.js#L239). As you can see, it affects `[4]` and `[5]`, which are `tx` and `ty` respectively. However, according to the wrong comment, `tx` is `[3]` and `ty` is `[5]`.

It's supposed to be
```
/**
 * 2x3 Matrix
 * @module mat2d
 *
 * @description
 * A mat2d contains six elements defined as:
 * <pre>
 * [a, b, c,
 *  d, tx, ty]
 * </pre>
 * This is a short form for the 3x3 matrix:
 * <pre>
 * [a, b, c,
 *  d, tx, ty,
 *  0, 0, 1]
 * </pre>
 * The last row is ignored so the array is shorter and operations are faster.
 */
```

baseten:
OK, great, that's what I thought

stefnotch:
Oh, yes! I made a mistake, was really careless when writing down the 3x3 matrix. Thanks for catching it.

stefnotch:
Fixed by #361

PR #361, opened 2019-05-10
correct order in mat2d doc comment

baseten:
The order of the elements in the description comment in `mat2d` was incorrect as discussed in https://github.com/toji/gl-matrix/issues/360

### DIFF
--- a/src/mat2d.js
+++ b/src/mat2d.js
@@ -7,16 +7,16 @@ import * as glMatrix from "./common.js";
  * @description
  * A mat2d contains six elements defined as:
  * <pre>
- * [a, c, tx,
- *  b, d, ty]
+ * [a, b, c,
+ *  d, tx, ty]
  * </pre>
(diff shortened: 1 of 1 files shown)


stefnotch:
Thank you! 🎉

src/mat2d.js
```diff
@@ -7,16 +7,16 @@ import * as glMatrix from "./common.js";
  * @description
  * A mat2d contains six elements defined as:
  * <pre>
- * [a, c, tx,
- *  b, d, ty]
+ * [a, b, c,
+ *  d, tx, ty]
```
magcius: This should probably be written as
```
[a, b,
 c, d,
 tx, ty]
```

My brain initially read it the wrong way because of the shape.

stefnotch: Good point. I fixed it in https://github.com/toji/gl-matrix/commit/120fe1057b510d4a49c3bcaffeb2b8cd52dd0cc0

PR #362, opened 2019-08-01
mat4: Support infinite far planes in mat4.frustum

magcius:
This is supported in mat4.perspective, so why not mat4.frustum?


stefnotch:
Mind writing a unit test for that?

magcius:
I probably won't have time for a bit to write one.

PR #363, opened 2019-08-18
Add scalar power for quaternions

stygianguest:


stefnotch:
Thanks for the great PR! It has some apparently rather useful functions complete with unit tests.

PR #364, opened 2019-08-18
Implement angular distance for quaternions

### LINKED ISSUE #336: Quaternion Distance
I'd like to take two quaternions and get a distance value which is the delta between their separate orientations. Any methods to do this I'm overlooking?

stygianguest:
Fixes #336


stefnotch:
Thank you!

kevzettler:
great addition!

PR #367, opened 2019-09-05
Update docs to express angles are in radians

mix3d:
No functionality changes, but some function inputs were renamed to match the usage in other classes, and the jsdoc updated to match. 

I lost an hour of debugging because it wasn't explicit in the docs that vec2/vec3 rotations were in radians, hopefully this prevents others from the same problem.

Also cleaned up some messy whitespace.


stefnotch:
Good point,  gl-matrix's documentation and function parameter names were a bit inconsistent here. Thanks for the PR, merged! 😄 

(The Travis CI failure seems to be unrelated.)

PR #376, opened 2020-01-19
Fix missing element in Frobenius norm for mat4

shmookey:


stefnotch:
Thank you for the PR. I'm sorry for not having caught that issue sooner.

PR #382, opened 2020-03-29
Set correct type for setMatrixArrayType

dcbrwn:
Typescript complained about unknown "Type". Judging from the code `Float32ArrayConstructor | ArrayConstructor` is appropriate.


stefnotch:
Thank you very much!

dcbrwn:
wow! that was quick!

PR #384, opened 2020-04-07
Improve mat4 determinant algorithm

rhendric:
This version shaves off two multiplication operations.


stefnotch:
Thank you very much!

PR #385, opened 2020-04-08
Improve mat4 adjoint algorithm

rhendric:
This is copied from the invert algorithm, which was obviously faster.


stefnotch:
Thank you very much!

Issue #386, opened 2020-04-15
Proposal: Relax type to arrays in types.d.ts 


anvaka:
Hello,

I was curious if community would be open to relax type strictness in the [types.d.ts](https://github.com/toji/gl-matrix/blob/master/src/types.d.ts) file. 

Instead of 

``` ts
declare type vec2 = [number, number] | Float32Array;
```

I propose to use

``` ts
interface IndexedCollection {
    readonly length: number;
    [n: number]: number;
}
declare type vec2 = [number, number] | IndexedCollection;
```

Main motivation behind this is that it is possible to to change matrix type to be, say `Float64Array` or anything else that looks like an array via `setMatrixArrayType()`, and current type definitions do not reflect this accurately.

If you are open to this suggestion I'd be happy to send a pull request.

stefnotch:
I'm all for it. After all, WebGL also accepts other typed arrays. So, yes, please do open a PR.

PR #387, opened 2020-04-15
Changed array type

anvaka:
As discussed in https://github.com/toji/gl-matrix/issues/386

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -1,22 +1,27 @@
+interface IndexedCollection {
+    readonly length: number;
+    [index: number]: number;
+}
+
 // prettier-ignore
 declare type mat2 =
   | [number, number, 
      number, number]
-  | Float32Array;
+  | IndexedCollection;
 
 // prettier-ignore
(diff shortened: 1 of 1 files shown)


src/types.d.ts
```diff
@@ -1,4 +1,4 @@
-interface IndexedCollection {
+interface IndexedCollection extends Iterable<number> {
```
stefnotch, line 1: What's the use case for making it iterable?

anvaka: Without it passing mat4/vec4 objects to `WebGLRenderingContext.uniformMatrix[234]fv()` fails. See [here](https://github.com/microsoft/TypeScript/blob/deb5bac520a1134e188fc6f01fa4f52b66d2899b/lib/lib.dom.iterable.d.ts#L332-L340).

stefnotch: Thank you for thoroughly testing everything!

anvaka:
Thank you for the merge! If there are any other issues discovered - I'd be glad to help

PR #402, opened 2020-08-27
add combined TRS decompose for mat4

### LINKED ISSUE #401: combined TRS decompose for mat4
(unrelated to #305)

`mat4.fromRotationTranslationScale` exists but not the other way round (i.e. [`Matrix4.decompose`](https://github.com/mrdoob/three.js/blob/4020fd0f74e457e012a1cc73e5ab1fbf6d2eb137/src/math/Matrix4.js#L719) from Three.js)

The `getX` functions from #204 are excellent, but if both rotation and scale are desired there is a wasted call to `getScaling` (position is independent)

I can do a PR but the only thing I am unsure of is the function signature, since I can't find an example of multiple `out`s on a function (besides `quat.getAxisAngle` which sets the `vec3` and returns the angle). I am also not sure if there was a concious decison to exclude it because of this.

randName:
closes #401 . I settled on returning only the rotation quaternion.

I added some tests which are basically copies of the other `getX` functions, but let me know if some are spurious or more are required.

### DIFF
--- a/src/mat4.js
+++ b/src/mat4.js
@@ -1227,6 +1227,65 @@ export function getRotation(out, mat) {
 }
 
 /**
+ * Decomposes a transformation matrix into its rotation, translation
+ * and scale components. Returns only the rotation component
+ * @param  {quat} out_r Quaternion to receive the rotation component
+ * @param  {vec3} out_t Vector to receive the translation vector
+ * @param  {vec3} out_s Vector to receive the scaling factor
+ * @param  {ReadonlyMat4} mat Matrix to be decomposed (input)
+ * @returns {quat} out_r
+ */
+export function decompose(out_r, out_t, out_s, mat) {
+  getTranslation(out_t, mat);
+  getScaling(out_s, mat);
+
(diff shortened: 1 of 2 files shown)


src/mat4.js
```diff
@@ -1226,6 +1226,65 @@ export function getRotation(out, mat) {
...
+ * and scale components. Returns only the rotation component
+ * @param  {quat} out_r Quaternion to receive the rotation component
+ * @param  {vec3} out_t Vector to receive the translation vector
+ * @param  {vec3} out_s Vector to receive the scaling factor
+ * @param  {ReadonlyMat4} mat Matrix to be decomposed (input)
+ * @returns {quat} out_r
+ */
+export function decompose(out_r, out_t, out_s, mat) {
+  getTranslation(out_t, mat);
+  getScaling(out_s, mat);
```
stefnotch, line 1240: Looks good! Though, one minor performance improvement would be to manually inline those two functions.

stefnotch:
Looks good, thank you very much!

Issue #403, opened 2020-09-29
Documentation: Clarify quat.fromEuler rotation order


gkjohnson:
The documentation for `quat.fromEuler` does not specify what order the euler angles are applied in. Referencing [three.js' implementation](https://github.com/mrdoob/three.js/blob/master/src/math/Quaternion.js#L235-L240) and [looking at the quat code](https://github.com/toji/gl-matrix/blob/master/src/quat.js#L472-L475) it seems that they are applied in ZYX order using intrinsic rotations. It would be great to add that information to the docs -- I'd assumed XYZ intrinsic.

Semi related but it would be nice to be explicit about the order that the quaternion is laid out in, as well, with elements 0, 1, 2 storing the complex xyz components while w is the real component in element 3. I suppose that's common but I've come across some other conventions that store the real component in element 0 so you never know!

And thanks for the awesome library!

stefnotch:
That's true, the `quat.fromEuler` is a bit confusing and should probably be changed to be in line with other libraries.

(Relevant issue about the inverse function: https://github.com/toji/gl-matrix/issues/329 )

And yes, the fact that the quaternion order wasn't documented is a minor oversight. Funnily enough, it is documented for dual quaternions.

DCtheTall:
@stefnotch I'd be willing to send a PR for adding documentation and/or changing the function to be more like other libraries if you want that. I am aware it's a breaking change so if you just want to document it that works for me 😄

gkjohnson:
A bit outside the scope of this issue but it would be nice to be able to specify the rotation order when converting euler to quaternion:

```js
quat.fromEuler( quat, x, y, z, 'xyz' );
```

It could default to `zyx` for the sake of backwards compatibility too.

stefnotch:
After thinking about this for a while, I suppose encoding that information right into the function signature would be a decent idea. For example, DirectXMath has a function named `XMQuaternionRotationRollPitchYawFromVector`. I'd propose a much shorter name like `fromEulerXYZ`. 

Also, for some reason, the `XZY` order is also somewhat popular (e.g. DirectXMath, Unity, SharpDX, ...). Not sure why though.

gkjohnson:
> Also, for some reason, the XZY order is also somewhat popular (e.g. DirectXMath, Unity, SharpDX, ...). Not sure why though.

[...]

Issue #369, opened 2019-11-05
Handle projection matrices for WebGPU-style depth ranges


magcius:
See: immersive-web/webxr#894, gpuweb/gpuweb#416

Currently, gl-matrix is modelled after GLM, which uses a clip space defined with a Z range of -1...1. This is in contrast with other graphics APIs, including Direct3D 12, Vulkan, Metal, and WebGPU, which all use a depth range of 0...1.

This could cause an issue when users naively port their WebGL-using code to WebGPU. It will appear like the near plane is much further in than it is, as anything less than 0 will be clipped.

It might make sense for functions like mat4.perspective, mat4.frustum, mat4.ortho, etc. to all support a clip-space of 0...1, either as an argument, a global setting (e.g. `mat4.useWebGPUClipSpace = true`), separate functions (e.g. `mat4.frustumWebGPU()`), or as functions to convert between the two for a typical projection matrix. For the lattermost, see this, which I have implemented in my project: https://github.com/magcius/noclip.website/blob/master/src/gfx/helpers/ProjectionHelpers.ts

GLM handles these with preprocessor defines, which are not the best choice.

At the same time, it might be interesting to think about designing an API to handle reverse-depth projection matrices as well.

Thoughts?

stefnotch:
I fully agree that gl-matrix should support this. Now, my thoughts as to how that support should look:

> global setting

I think that'd be a footgun. The current global settings do not significantly alter the results of the calculations, a depth range global setting on the other hand would.

> argument

How hard would it be to be able to specify the depth range as an argument? If this option is chosen, there is another choice that has to be made, which is whether gl-matrix should break backwards compatibility or have default values for the depth range.

> separate functions 

This is a pretty simple option that won't break backwards compatibility. 

> functions to convert between the two 

This would be a quite neat option, but it's not as performant as the other ones.

magcius:
> I think that'd be a footgun. The current global settings do not significantly alter the results of the calculations, a depth range global setting on the other hand would.

I mention it because it's "easy" (set it once and forget it), and it's what GLM did.

> How hard would it be to be able to specify the depth range as an argument? If this option is chosen, there is another choice that has to be made, which is whether gl-matrix should break backwards compatibility or have default values for the depth range.

I think it might make sense to have a flags argument, e.g.
```ts
mat4.frustum(dst, left, right, top, bottom, near, far, mat4.REVERSED_DEPTH | mat4.Z_RANGE_0_1);
```

[...]

PR #407, opened 2020-10-06
Support multiple intrinsic orders in quat.fromEuler

DCtheTall:
Fixes [#403](https://github.com/toji/gl-matrix/issues/403).


DCtheTall:
@stefnotch please review this change when you get a chance 😄

src/quat.js
```diff
@@ -447,36 +447,50 @@ export function fromMat3(out, m) {
...
- * @param {x} Angle to rotate around X axis in degrees.
- * @param {y} Angle to rotate around Y axis in degrees.
  * @param {z} Angle to rotate around Z axis in degrees.
+ * @param {y} Angle to rotate around Y axis in degrees.
+ * @param {x} Angle to rotate around X axis in degrees.
  * @returns {quat} out
  * @function
  */
-export function fromEuler(out, x, y, z) {
+export function fromEuler(out, z, y, x) {
```
gkjohnson, line 459: I don't think I agree with changing the argument order -- I see you've changed the contents of the function so the result is the same but I feel changing the argument order to something other than x y and z is confusing. Euler angles are often stored in XYZ order independent of application order in my experience. I think all that should change is the documentation.

src/quat.js
```diff
@@ -447,36 +447,50 @@ export function fromMat3(out, m) {
 }
 
 /**
- * Creates a quaternion from the given euler angle x, y, z.
+ * Creates a quaternion from the given euler angle z, y, x.
```
gkjohnson, line 450: In my opinion the docs should be explicit about the order that rotations are applied in. I think simply listing them here just looks like it's listing the argument order. It's also important to specify whether the convention is using [intrinsic or extrinsic rotations](https://en.wikipedia.org/wiki/Euler_angles#Chained_rotations_equivalence) regarding the order listed.

From my brief testing (please someone else verify -- I'd hate to be the one that causes some misdocumentation 😁) it seems the angles are applied in ZYX intrinsic order which is the same as XYZ extrinsic order. Either specification works for me.

src/quat.js
```diff
@@ -447,36 +447,50 @@ export function fromMat3(out, m) {
...
+ * @param {quat} out the receiving quaternion
+ * @param {x} Angle to rotate around X axis in degrees.
+ * @param {y} Angle to rotate around Y axis in degrees.
+ * @param {z} Angle to rotate around Z axis in degrees.
+ * @returns {quat} out
+ * @function
+ */
+export function fromEulerXYZ(out, x, y, z) {
+  return fromEuler(out, z, y, x);
+}
```
gkjohnson, line 492: Just swapping the order in which angles are passed into the function does not change the rotation order -- it actually yields a different rotation because the angles are being rotated about the wrong axes now. Three.js [uses different formulas for the quaternion](https://github.com/mrdoob/three.js/blob/master/src/math/Quaternion.js#L214-L254) when applying euler angles in different orders.

stefnotch:
gkjohnson has left some pretty good comments, please do take a look at them.  And, the order should probably be `XYZ intrinsic`, just because that's what three.js does by default.

src/quat.js
DCtheTall: I see now that I misinterpreted the original issue. I spent some more time reading up on Euler angles (it's been a couple years since my computer graphics class 😅 ).

> I think all that should change is the documentation.

Done, though I did refactor the internals of both into a private function to avoid duplicated code.

src/quat.js
DCtheTall: > Just swapping the order in which angles are passed into the function does not change the rotation order

Yep, I see my mistake now. Changed to the new formula.

> Three.js supports changing rotation order and uses different formulas for the quaternion when applying euler angles in different orders.

Yep, looks like they're using intrinsic ordering labels too.

src/quat.js
DCtheTall: > In my opinion the docs should be explicit about the order that rotations are applied in.

Done.

> [I]t seems the angles are applied in ZYX intrinsic order [are] the same as XYZ extrinsic order

[This article seems to agree](https://dreamanddead.github.io/2019/04/24/understanding-euler-angles.html) with a proof. The proof can also be trivially modified to show that XYZ intrinsic order is equivalent to ZYX extrinsic order as well if you just change the subscripts.

DCtheTall:
@gkjohnson I addressed your review comments, let me know if this is more what you had in mind.

gkjohnson:
@DCtheTall Great thanks! A few thoughts:

- It's up to the maintainers of this repo but given that speed is such a big push for `gl-matrix` I suspect that having an internal function that uses a switch case to select which order to use might be undesirable (though I know I suggested in the other issue 😬) If it _is_ okay then I think a single `fromEuler` function with an optional "order" argument that specifies some documented intrinsic or extrinsic order and defaults to the current behavior is cleaner. Either way I think it should just be one or the other.

- Given that an intrinsic rotation order is just the reverse of an extrinsic rotation (intrinsic XYZ rotation is the same as an extrinsic ZYX rotation) I don't think it's necessary to specify "Intrinsic" or "Extrinsic" in the function names. The library should just document what is being used per function and be consistent about it. If `gl-matrix` uses intrinsic XYZ order then a user who needs extrinsic XYZ order would just use the `getEulerZYX` method (assuming there are multiple methods).

Of course @stefnotch will have the final opinion on this but those are my 2 cents. Thanks for the improvements!

DCtheTall:
@gkjohnson thank you for putting so much thought into this. I am also open to adding a new argument to `fromEuler` with a default value, since that is a simpler change, though all we'd be saving at that point is a function invocation since then the `switch` would be moved to the other function body.

I'll defer to @stefnotch and see what his thoughts are.

stefnotch:
I suppose adding a new argument to 'fromEuler' would be reasonably clean. 

(Sort of related https://github.com/toji/gl-matrix/issues/369#issuecomment-553557148 )

DCtheTall:
@stefnotch 

> I suppose adding a new argument to 'fromEuler' would be reasonably clean.

Done.

Also decided to support all intrinsic orders while I was at it. I borrowed pretty shamelessly from Three.js's implementation but with the proper variable names for this library.

src/quat.js
```diff
@@ -447,32 +447,75 @@ export function fromMat3(out, m) {
...
 /**
- * Creates a quaternion from the given euler angle x, y, z.
+ * Creates a quaternion from the given euler angle x, y, z using the provided extrinsic order for the conversion.
  *
  * @param {quat} out the receiving quaternion
  * @param {x} Angle to rotate around X axis in degrees.
- * @param {y} Angle to rotate around Y axis in degrees.
  * @param {z} Angle to rotate around Z axis in degrees.
+ * @param {y} Angle to rotate around Y axis in degrees.
+ * @param {order} Intrinsic order for conversion, so far only 'xyz' and 'zyx' is supported.
```
gkjohnson, line 456: Looks like some of these comments are out of date now. The description says extrinsic order but the argument says intrinsic with only two options available and y is now listed as the third argument.

src/quat.js
```diff
@@ -447,32 +447,75 @@ export function fromMat3(out, m) {
...
+  let sy = Math.sin(y);
+  let cy = Math.cos(y);
 
-  out[0] = sx * cy * cz - cx * sy * sz;
-  out[1] = cx * sy * cz + sx * cy * sz;
-  out[2] = cx * cy * sz - sx * sy * cz;
-  out[3] = cx * cy * cz + sx * sy * sz;
+  if (typeof order != 'string') {
+    order = 'zyx';
+  }
```
gkjohnson, line 475: nit: because you're using `default` in the switch case this condition isn't needed.

src/quat.js
```diff
@@ -447,32 +447,75 @@ export function fromMat3(out, m) {
...
+  y *= halfToRad;
 
   let sx = Math.sin(x);
   let cx = Math.cos(x);
-  let sy = Math.sin(y);
-  let cy = Math.cos(y);
   let sz = Math.sin(z);
   let cz = Math.cos(z);
+  let sy = Math.sin(y);
+  let cy = Math.cos(y);
```
gkjohnson, line 471: nit: the variable order has been needlessly rearranged and no longer matches the argument order -- just something I noticed!

src/quat.js
DCtheTall: It's so `toLowerCase()` will not throw a TypeError.

src/quat.js
DCtheTall: Done.

src/quat.js
DCtheTall: Oops, thanks for catching that. Fixed.

DCtheTall:
@gkjohnson friendly ping

src/quat.js
```diff
@@ -447,20 +447,21 @@ export function fromMat3(out, m) {
...
 
 /**
- * Creates a quaternion from the given euler angle x, y, z.
+ * Creates a quaternion from the given euler angle x, y, z using the provided intrinsic order for the conversion.
  *
  * @param {quat} out the receiving quaternion
  * @param {x} Angle to rotate around X axis in degrees.
  * @param {y} Angle to rotate around Y axis in degrees.
  * @param {z} Angle to rotate around Z axis in degrees.
+ * @param {order} Intrinsic order for conversion, so far only 'xyz' and 'zyx' is supported.
```
gkjohnson, line 456: These docs are still out of date after adding support for the rest of the rotation orders specifically this part:

>  so far only 'xyz' and 'zyx' is supported.

It might also be worth noting that only ["Tait-Bryan angles"](https://en.wikipedia.org/wiki/Euler_angles#Chained_rotations_equivalence) are supported (as opposed to "Proper Euler angles" which specify a rotation about the same axis twice) which I saw the [three.js docs](https://threejs.org/docs/#api/en/math/Euler.order) clarify.

gkjohnson:
> @gkjohnson friendly ping

Thanks for the ping I didn't realize the ball was in my court. I just left a comment but after that's addressed I think the rest is up to the maintainers.

Thanks!

src/quat.js
gkjohnson: The default value for `order` might need to explicitly documented as well? I'm not sure if the docs generation process automatically detects inline parameter defaults.

DCtheTall: Done.

DCtheTall:
@stefnotch

stefnotch:
@DCtheTall Looks great, merged! Thank you and gkjohnson very much for all your work on this 🎉

stefnotch:
After taking a look at it, I decided to slightly tweak a few bits and bobs, mainly using the documentation comment to validate the parameters and defining the default angle order in common.js.
3a321ea79b4d29b836245bfff52569fe61d0e96a
0fc30ed8da8e168061be114f55f12410899082e6

dist/gl-matrix.js
```diff
@@ -5725,31 +5826,78 @@ THE SOFTWARE.
...
-    out[0] = sx * cy * cz - cx * sy * sz;
-    out[1] = cx * sy * cz + sx * cy * sz;
-    out[2] = cx * cy * sz - sx * sy * cz;
-    out[3] = cx * cy * cz + sx * sy * sz;
+
+    if (typeof order != 'string') {
+      order = 'zyx';
+    }
+
+    switch (order.toLowerCase()) {
```
magcius, line 5857: `toLowerCase()` causes an unavoidable string allocation in what is (possibly) a hot path. :/

I don't know if this is part of the API now, but I would prefer if people had to put their order argument in lower-case.

dist/gl-matrix.js
```diff
@@ -5725,31 +5826,78 @@ THE SOFTWARE.
...
+
+      case 'zxy':
+        out[0] = sx * cy * cz - cx * sy * sz;
+        out[1] = cx * sy * cz + sx * cy * sz;
+        out[2] = cx * cy * sz + sx * sy * cz;
+        out[3] = cx * cy * cz - sx * sy * sz;
+        break;
+
+      case 'zyx':
+      default:
```
magcius, line 5894: Falling back to `zyx` when specifying an order that isn't supported (e.g. `xyx`) feels like an API footgun, I would rather throw an error here.

stefnotch:
@magcius Those are some quite valid concerns. Luckily, this isn't a part of the official API yet, which means that a PR that changes this behaviour would be very welcome.

magcius:
I missed that these concerns were already addressed in https://github.com/toji/gl-matrix/commit/3a321ea79b4d29b836245bfff52569fe61d0e96a, apologies.

Issue #411, opened 2020-12-11
Deprecate and remove perspectiveFromFieldOfView


stefnotch:
See https://github.com/toji/gl-matrix/issues/410#issuecomment-743434933

PR #413, opened 2021-01-12
Add mat4.perspectiveZO, mat4.orthoZO for WebGPU

toji:
Putting this up as a pull request in case there's any feedback.

The upcoming WebGPU API uses Normalized Device Coordinates with a [0, 1] Z range, which differs from WebGL's [-1, 1] range. To get the best precision out of WebGPU pages will need to create perspective matrices a bit differently.

This change introduces the `mat4.perspectiveZO()`, which is identical in functionality to the previous `mat4.perspective()` method except it maps to the [0, -1] range. For consistency the previous method has been renamed to `mat4.perspectiveNO()` with an alias of `mat4.perspective()`. This could later allow users to configure the library to use one projection method or the other by default. Naming convention borrowed from the native [glm library](https://github.com/g-truc/glm) because I can't think of anything better that doesn't end up with 50 character long method names.

### DIFF
--- a/src/mat4.js
+++ b/src/mat4.js
@@ -1525,6 +1525,8 @@ export function frustum(out, left, right, bottom, top, near, far) {
 
 /**
  * Generates a perspective projection matrix with the given bounds.
+ * The near/far clip planes correspond to a normalized device coordinate Z range of [-1, 1],
+ * which matches WebGL/OpenGL's clip volume.
  * Passing null/undefined/no value for far will generate infinite projection matrix.
  *
  * @param {mat4} out mat4 frustum matrix will be written into
@@ -1534,7 +1536,7 @@ export function frustum(out, left, right, bottom, top, near, far) {
  * @param {number} far Far bound of the frustum, can be null or Infinity
  * @returns {mat4} out
  */
-export function perspective(out, fovy, aspect, near, far) {
+export function perspectiveNO(out, fovy, aspect, near, far) {
   let f = 1.0 / Math.tan(fovy / 2),
     nf;
   out[0] = f / aspect;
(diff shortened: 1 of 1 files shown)


kainino0x (approved):
Is something similar needed for ortho() and perspectiveFromFieldOfView()?

toji:
perspectiveFieldOfView is planned to be [deprecated and removed](https://github.com/toji/gl-matrix/issues/411). You're right that ortho will need the same thing, though.

toji:
Added the ortho version.

src/mat4.js
```diff
@@ -1562,6 +1564,53 @@ export function perspective(out, fovy, aspect, near, far) {
...
+ * @param {mat4} out mat4 frustum matrix will be written into
+ * @param {number} fovy Vertical field of view in radians
+ * @param {number} aspect Aspect ratio. typically viewport width/height
+ * @param {number} near Near bound of the frustum
+ * @param {number} far Far bound of the frustum, can be null or Infinity
+ * @returns {mat4} out
+ */
+export function perspectiveZO(out, fovy, aspect, near, far) {
+  let f = 1.0 / Math.tan(fovy / 2),
+    nf;
```
kainino0x, line 1588: Any reason not to `let nf` inside the `if()` where it's used?

kainino0x (approved):
Should I review the math in detail? Otherwise, LGTM

src/mat4.js
kainino0x: `const nf`, actually

magcius:
You missed `frustumZO`, and I thought we were going to take a flag approach as detailed in https://github.com/toji/gl-matrix/issues/369#issuecomment-553557148 so we could also handle reverse depth at the same time?

PR #418, opened 2021-03-19
fix!: use dot product to compute quaternions equality

### LINKED ISSUE #292: Quaternion Equality
Quaternions are also considered equal when all signs are flipped. i.e.
`[x,y,z,w] == [-x,-y,-z,-w]`

Should the equals function take this into account? Or, should normalizing the quaternions take this into account?

Wawa27:
The old method for quaternions equality returns false with flipped orientation (eg [0, 0, 0, 1] and [0, 0, 0, -1])
This pull request uses dot product instead, assuming quaternions are unit length

It is breaking change because old checks for equality when quaternions are not unit length won't work anymore

fixes #292

### DIFF
--- a/spec/gl-matrix/quat-spec.js
+++ b/spec/gl-matrix/quat-spec.js
@@ -776,20 +776,23 @@ describe("quat", function() {
     });
 
     describe("equals", function() {
-        let quatC, quatD, r0, r1, r2;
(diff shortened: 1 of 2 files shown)


stefnotch:
Looks great, thank you very much!

PR #419, opened 2021-03-20
Update mat4.js.html

classAndrew:
https://glmatrix.net/docs/module-mat4.html#.getScaling
Changes 'paramter' to 'parameter'.
There is no useful functionality. This is just a spelling mistake in the docs.


stefnotch:
Should probably be fixed in the source instead, since the docs are autogenerated https://github.com/toji/gl-matrix/blob/be9e9b5e5ba6603cd7e06b868696a241d003338e/src/mat4.js#L1145

Issue #420, opened 2021-03-26
[AssemblyScript] gl-matrix port to WebAssembly


StEvUgnIn:
Hi everyone,
I am working since yesterday on the port of gl-matrix to WebAssembly. 

Define objectives are:
- [X] compatibility with TypeScript preserved (using `std/assembly/assemblyscript` definitions)
- [ ] minimal changes from the original code
- [ ] encapsulation and compatibility with projects using the JavaScript implementation

I will publish as soon as I have a solid draft to present :) But the code of `gl-matrix` is concise, not too shaby for JavaScript code, easily portable to TS/AS environments.

StEvUgnIn:
Don't hesitate to share your feedback on my ongoing work here

stefnotch:
Out of curiosity, how will garbage collection be handled?

StEvUgnIn:
Garbage collection makes more sense in bigger projects. Memory is handled manually on Webassembly by the programmer. You have a ton of projects porting C/C++, JavaScript, Rust to Webassembly. 

Webassembly is designed to work alongside JavaScript. That means that the binary needs glue code to be able to run in an application (client or server).

trusktr:
AssemblyScript has automatic GC currently.

- if an AssemblyScript Wasm module is no longer referenced by JS, the it is collected (this is regular stuff, not assemblyscript-specific).
- If an AssemblyScript Wasm module is still held by JS, then every time JS invokes the module, AssemblyScript does it's own GC (custom implementation, not the upcoming Wasm GC) on each call. The developer, by default, does not need to worry about it.

StEvUgnIn:
Thanks for your explanation @trusktr 🙂

StEvUgnIn:
https://travis-ci.org/github/toji/gl-matrix/builds/764734352 😎

StEvUgnIn:
@stefnotch @trusktr @pyrotechnick @AlexMax 
https://github.com/toji/gl-matrix/pull/421#pullrequestreview-626668492

psnet:
I tested on another project same idea for calculations on WASM and due to the bridge JS - WASM where non-scalar values are copied between memory of JS and WASM so it is actually slower than done in pure JS.

So this entire idea is not efficient

PaperPrototype:
[...]

PR #421, opened 2021-03-27
[AssemblyScript] porting source from JS to AS

StEvUgnIn:
Follows #420


StEvUgnIn:
Next steps are: 
- [ ] Fix typings between JS and AS/TS
- [ ] Port JS spec to AS/TS spec (spec.ts)
- [ ] Be able to work fully with emitted JS from AS implementation
- [ ] Write an encapsulated interface layer for using WASM from any web programming environment
- [ ] Benchmark performances between JS and AS modes

Please comment here if you are interested to generously help.

trusktr:
This will be neat for AS users to have!

Is there a particular reason to commit the build artefacts to the repo? Those can be built during publish time when publishing to NPM for example.

Why build Wasm modules? 

If people will import the JS code into their JS projects, or import the AS code into their AS projects, then the module builds would not be needed. The Wasm modules are needed only for JS users who would import the Wasm modules instead of the JS code. 

Wondering if that's it's worth building modules. Would the barrier between JS and Wasm allow this to be faster?

It would be sweet to see what perf diff looks like in a JS example that makes many calls into the module, vs plain JS and vs plain AS.

StEvUgnIn:
Hi @trusktr 

> This will be neat for AS users to have!
> 

Thanks! I hope we could see more WebGL frameworks emerge on the WebAssembly platform and this helps to train programming in Assemblyscript 🙂

> Is there a particular reason to commit the build artefacts to the repo? Those can be built during publish time when publishing to NPM for example.

I was unsure at first, but I decided to include this because of the AssemblyScript conventions and I suppose someone could just download the file to include it in an existing project, especially during this testing phase. 

> 
> Why build Wasm modules?
> 
> If people will import the JS code into their JS projects, or import the AS code into their AS projects, then the module builds would not be needed. The Wasm modules are needed only for JS users who would import the Wasm modules instead of the JS code.
> 

Yes, exactly.

A default `.gitignore` lies in the `build` repository which actually lets you keep the built modules by default.

> Wondering if that's it's worth building modules. Would the barrier between JS and Wasm allow this to be faster?
> 
> It would be sweet to see what perf diff looks like in a JS example that makes many calls into the module, vs plain JS and vs plain AS.

I am looking forward to try both implementations from the browser 😃

asconfig.json
```diff
@@ -0,0 +1,20 @@
...
+      "debug": true
+    },
+    "release": {
+      "binaryFile": "build/optimized.wasm",
+      "textFile": "build/optimized.wat",
+      "sourceMap": true,
+      "optimizeLevel": 3,
+      "shrinkLevel": 1,
+      "converge": false,
+      "noAssert": false
```
MaxGraey, line 16:
```suggestion
      "shrinkLevel": 0,
      "converge": false,
      "noAssert": true
```
if you required best perf

StEvUgnIn: Thanks! @MaxGraey

StEvUgnIn:
I have a weird compiler output due to type declared inside modules yet..
https://gist.github.com/StEvUgnIn/acbfed089a24b1c671f79b1a36723a06

But so far, it seems to go well.

StEvUgnIn:
image
JavaScript output now renders

StEvUgnIn (commented):
This work finalizes the first build targeting Javascript and Webassembly.

StEvUgnIn (commented):
msedge_BdF7kMQpT4

loader/index.js
```diff
@@ -0,0 +1,14 @@
+import wasm from '../build/optimized.wasm';
+
+let modules;
+
+wasm({ ...imports }).then(instance => {
```
StEvUgnIn, line 5: This is temporary, before I implement a better solution for exporting WebAssembly modules. Solutions are either:

- Complying on @assemblyscript/loader using `loader.instantiateStreaming`
- Making a WASM file for each module (curently max size for synchronous compilation is 4KB). Current WASM file is >100KB with `-O3` optimization
- Develop a processor script to pass `<script>` to WebAssembly using `eval` and `<noscript>`

StEvUgnIn: ### [WebAssembly.Module.exports()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Module)

StEvUgnIn: I'm currently experimenting with [as-bind](https://github.com/torch2424/as-bind)

mrchantey:
## 32 Bit Support
This PR is great! I'm in the process of implementing the brach in my project now. 
I'd just like to flag the 64 bit operations. We may want to have flexibility, for either allowing users to make the call at runtime, or exporting two seperate builds.
For starters we could declare something like the following in `common.ts` and then use it throughout the library.
```ts
//64 
export type Float = f64
export type FloatArray = Float64Array
export const FloatMath = Math

//32
export type Float = f32
export type FloatArray = Float32Array
export const FloatMath = Mathf
```

StEvUgnIn:
Hello @mrchantey,

Please you are welcome for trying and sharing your proposition.

StEvUgnIn:
I didn't make any advancement on this PR. I was waiting for `as-bind` to support Array/Float32Array/Float64Array before going any further.

psnet:
I tested on another project same idea for calculations on WASM and due to the bridge JS - WASM where non-scalar values are copied between memory of JS and WASM so it is actually slower than done in pure JS.

So this entire idea is not efficient

StEvUgnIn:
@psnet Thank you for your feedback. I stopped working on this port a few years back already, but I am keeping an eye on ongoing discussions.

PR #430, opened 2021-06-02
Optimize angle(v1, v2)

munsocket:
Math.sqrt() is heavy operation and can be used only once


stefnotch:
Good point, thanks for this 👍

munsocket:
That was fast, thanks! Found another one for Vec2.

PR #432, opened 2021-06-02
Convert to type "module" and use package exports

### LINKED ISSUE #425: Migrate to fully ESM
https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77

I personally would prefer option one from the linked blog post. After all, gl-matrix is at the point where barely anything ever changes, which means that pointing people with old build tools to an older version of gl-matrix should be fine. We'd have to release a new major version of gl-matrix though.

> Migrating
> There are two ways to move your packages to ESM:
> 1.   Pure ESM
>   This has the benefit that it’s easier to set up. You just add "type": "module" to your package.json, require Node.js 12, update docs & code examples, and do a major release.
> -- Option one

lohfu:
This PR uses `type: "module"` and package exports to fully support ESM in modern versions of Node.

Closes #425


stefnotch:
I personally would prefer option one from the linked blog post. After all, gl-matrix is at the point where barely anything ever changes, which means that pointing people with old build tools to an older version of gl-matrix should be fine. We'd have to release a new major version of gl-matrix though.

> Migrating
> There are two ways to move your packages to ESM:
> 1.   Pure ESM
>   This has the benefit that it’s easier to set up. You just add "type": "module" to your package.json, require Node.js 12, update docs & code examples, and do a major release.
> -- Option one

lohfu:
This is completely fine by me. The only caveat is that this not only requires newer versions of node, but to also be running Node with ESM files. Since TypeScript currently doesn't support outputting Node compatible ESM files (they do not include file extensions in import calls) the new gl-matrix wont be consumable through typescript in Node. This should not be a problem in frontend code though, which is the main use case of gl-matrix.

stefnotch:
Good to know about that limitation. I suppose this should be fine, like you've said.

lohfu:
The project has now been converted to `type: "module"` and the CJS and nested package.json builds have been removed. I also converted all build and config files to ESM. I had to update rollup to support this fully.

lohfu:
The Travis build seems to be using Node 12.16.1, which does not support ESM modules. All tests pass locally (in ^12.17, ^13.2 and >= 14).

stefnotch:
Feel free to bump up the Node version used in Travis https://github.com/toji/gl-matrix/blob/6f5c321b694d747a16e0aaafe0a9c5d39451985a/.travis.yml#L4

lohfu:
ok sorted, set it to latest LTS 14.17.0.

lohfu:
should the common file be exposed in a export? for something like `import { toRadian } from 'gl-matrix/common'`?

stefnotch:
Yes please, do expose the common file. The toRadian and other methods are useful to have.

lohfu:
done! i see now common is exported as `glMatrix` from `index.js`, not `common`... should this be renamed, or would that be too much of a breaking change?

stefnotch:
Oh, if it's already being exposed as `glMatrix`, we can keep it that way.

lohfu:
it's exposed in a little different way... today you can:

```js
import { glMatrix } from 'gl-matrix'

glMatrix.toRadian()
```

with my change you can also:

```js
import { toRadian } from 'gl-matrix/common'

toRadian()
```

i was wondering if i you think i should rename the named export `glMatrix` > `common`, rename the package export `common` > `glMatrix` or leave it.

stefnotch:
I think it's currently fine with `import { glMatrix } from 'gl-matrix'`. If both options are now possible, that's also fine.

lohfu:
ok, then I'm feeling pretty done with this PR and it should be ready for merge!

stefnotch:
Thank you!

birkskyum:
It's still not full esm - flagged as FAUX by https://node-modules.dev/ , rather than CJS / ESM

Screenshot 2025-09-29 at 13 35 59

It needs this
- #455

PR #437, opened 2021-09-10
random should only return unit vector when scale is undefined

marcusround:
I encountered unexpected behaviour when passing dynamic values to the `scale` parameter of `vec2.random`.

calling `random(vec, 0)` should return a vector of length `0`, not `1`.


stefnotch:
Thank you! I definitely prefer this over the previous behavior

PR #448, opened 2022-06-07
use symmetric half up rounding in all vector modules. fixes #447

### LINKED ISSUE #447: vec*.round(out, a) is asymmetric
little known fact: javascript's `Math.round(n)` is asymmetric; it will round negative numbers towards +∞:

```javascript
console.log(Math.round(1.5))     // 2    probably we expected this
console.log(Math.round(-1.5))    // -1   what the fack?
```

because all of our vector modules rely on the built-in `Math.round` we get this probably unexpected behavior:
```javascript
const out = vec2.create()
console.log(vec2.round(out, [-1.5, -1.5 ]))  // out === [ -1, -1 ]
```

This is probably _not_ what most people are expecting. In my own games/simulations I use this module https://www.npmjs.com/package/round-half-up-symmetric to fix this behavior.  

Would you be open to a PR that leverages this module wherever we use `Math.round(n)` ?

mreinstein:


toji (approved):
Seems like a reasonable change, and I agree that it probably matches developer expectations better.

PR #449, opened 2022-06-09
Replace Math.hypot with Math.sqrt

### LINKED ISSUE #446: Math.hypot is really slow
Math.hypot is painfully slow on every modern javascript environment. To live up to the project's goal: "glMatrix is designed to perform vector and matrix operations stupidly fast" I suggest changing use of `Math.hypot(x, y)` to `Math.sqrt(x * x + y * y)` for significantly improved performance.

Is there an argument for continuing to use hypot? It is more accurate for very small and very large numbers (eg- `1e200`) [[1]](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot#description). But I'm not sure those edge cases matter for a WebGL library that is meant to be _fast_.

If this is something that would be considered, I am happy to make a Pull Request with the change. Would love to hear your thoughts.

Benchmark results of hypot vs sqrt (smaller is better):

| environment | hypot | sqrt |
|-------------|-------|------|
| chrome      | 235 ms | 12 ms |
| firefox     | 64 ms  | 12 ms |
| safari      | 390 ms | 74 ms |
| node        | 248 ms | 11 ms |

<details>
<summary>test script</summary>

```js
const size = 10000000
let sum = 0
let x = 1.1
const y = 2.2
const z = 3.3

console.time("hypot loop")
for (let i = 0; i < size; i++) {
  sum += Math.hypot(x, y, z)
  x += 1
}
console.timeEnd("hypot loop")

console.time("sqrt loop")
sum = 0
x = 1.1
for (let i = 0; i < size; i++) {
  sum += Math.sqrt(x * x + y * y + z * z)
  x += 1
}
console.timeEnd("sqrt loop")
```
</details>

platypii:
Fixes #446 

This PR replaces all instances of `Math.hypot` with `Math.sqrt`. It also removes the polyfill for hypot on really old browsers.

The reason is that `sqrt` is MUCH faster than `hypot` on all modern javascript engines. See #446 for some benchmarks.

While hypot is nicer to write, and is technically more accurate especially for very large and very small numbers, I think that the performance gains from switching to `sqrt` are worth it. Especially for a library with the goal:

> glMatrix is designed to perform vector and matrix operations stupidly fast!

Suggestions and feedback appreciated.

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -55,11 +55,3 @@ export function toRadian(a) {
 export function equals(a, b) {
   return Math.abs(a - b) <= EPSILON * Math.max(1.0, Math.abs(a), Math.abs(b));
 }
-
-if (!Math.hypot)
-  Math.hypot = function () {
-    var y = 0,
-      i = arguments.length;
-    while (i--) y += arguments[i] * arguments[i];
-    return Math.sqrt(y);
-  };
--- a/src/mat2.js
+++ b/src/mat2.js
@@ -307,7 +307,7 @@ export function str(a) {
  * @returns {Number} Frobenius norm
  */
 export function frob(a) {
-  return Math.hypot(a[0], a[1], a[2], a[3]);
+  return Math.sqrt(a[0] * a[0] + a[1] * a[1] + a[2] * a[2] + a[3] * a[3]);
 }
 
 /**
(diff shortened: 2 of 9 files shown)


mreinstein:
very elegant solution @platypii

PR #457, opened 2023-05-14
Fixed mat4.lookAt and mat4.targetTo

L1quidH2O:
Fixed mat4.lookAt and targetTo: when the direction and up vectors are parallel, it returned an incorrect rotation because the crossproduct returned a 0 vector, now it shifts the direction a bit to get a more accurate result.

Also seems its been years since the project as been built so done that too.


psnet:
It seems that this is fix for the issue https://github.com/toji/gl-matrix/issues/466
but fix is bit large, all need to do is compare
```ts
if (position[0] === target[0] && position[2] === target[2]) {
  target[2] += 0.00000001;
}
```

PR #469, opened 2024-05-13
Feature: For `common.glMatrix` `equals` method added optional param `tolerance` by default equals to `EPSILON`

psnet:


toji:
Thank you for your contribution!

PR #475, opened 2024-10-06
Align docs with parameters names

HarelM:
This is to align parameters names to comments.

I'm using typedoc which complains about misalignment between docs comments and actual code.
I've fixed what it found (although the version I'm using is not from latest code).
Regardless, these fixes are correct.

It would be great if you could also release a new version after merging this.
Thanks!


toji (approved):

toji:
Thanks!

HarelM:
Let me know when a new version is available so I can try out the docs generation to see if there are other issues.

PR #476, opened 2024-10-11
Optimized quaternion rotation routine

infusion:
Hello, 

I'm the author of Quaternion.js and derived a simpler method to rotate a vector using normalized quaternions, reducing the number of multiplications quite a bit. Since gl-matrix is developed with performance in mind, I think this addition can improve quite a few code bases.

All your unit tests are passing. I removed the performance test link, since jsPerf is offline for quite some time now.

Robert

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2015-2021, Brandon Jones, Colin MacKenzie IV.
+Copyright (c) 2015-2024, Brandon Jones, Colin MacKenzie IV.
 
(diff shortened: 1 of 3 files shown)


gene-ressler:
This looks like an excellent improvement. I wish it would be merged. My only nit is that adding two floats to implement multiplication by two is probably not worthwhile and possibly slower. At any rate, JIT compilation will do the right thing if the code ends up in a critical place.

toji:
Thank you for the contribution, and sorry it took a while to merge!

PR #479, opened 2024-11-13
Fixes typos documentation comments in mat4.js

LordArugula:
In the doc comments for the `fromRotationTranslation...` functions, `quat4` is used as the type instead of `quat`. See https://github.com/toji/gl-matrix/blob/master/src/mat4.js#L1048

Also fixes typo mentioned in https://github.com/toji/gl-matrix/pull/419


toji:
Thank you!

PR #480, opened 2024-12-16
Fix build errors by locking dependencies

hughsk-canva:
The project had stopped building successfully, including when installing locally. This appears to be due to a change in one of `jsdoc`'s dependencies that was breaking the generation of TypeScript declaration files.

This also includes a `package-lock.json` file to avoid these kind of transitory dependency issues in the future :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 tmp
 node_modules
 npm-debug.log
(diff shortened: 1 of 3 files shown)


toji:
Thank you!

PR #481, opened 2024-12-30
Fix return types of matrix invert functions

ddadamhooper:
Fix return types of matrix `invert()` functions to allow `null`.

Fixing the return type is important because it lets us catch errors in TypeScript before we ship them to customers.


toji:
Thank you!

PR #484, opened 2025-01-14
Run tests on push using GitHub Actions

hughsk-canva:
Following up on toji/gl-matrix#480, this adds a basic CI pipeline to the repository which runs tests on all branch pushes. A small change, but hopefully helpful :)

An example run is here, it should stick around for another week or two: https://github.com/hughsk-canva/gl-matrix/actions/runs/12759586150/job/35563603100

Screenshot 2025-01-14 at 13 09 45

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,16 @@
+name: Push
+on:
+  push:
+    branches: "*"
+
+jobs:
+  tests:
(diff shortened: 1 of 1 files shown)


toji:
Great! We'll give it a try!

hughsk-canva:
Great! Thanks @toji 🎉

PR #486, opened 2025-02-24
[glmatrix-next] Added type inference for static method return type

aidenfoxx:
Currently the static methods reduce the scope of the `out` variable on return to the associated `TypeLike` type. We can use type inference to guarantee the return type of the method matches the input. This enables chaining on static method return, or directly returning the method response. E.g:

```typescript
function doThing(): Vec3 {
  const test = new Vec3();
  return Vec3.transformMat3(test, ...).invert(test);
}
```

This also removed the need for type assertions in parts of the code.

### DIFF
--- a/src/quat2.ts
+++ b/src/quat2.ts
@@ -153,7 +153,7 @@ export class Quat2 extends Float32Array {
    * @param t - translation vector
    * @returns `out`
    */
-  static fromRotationTranslation(out: Quat2Like, q: Readonly<QuatLike>, t: Readonly<Vec3Like>): Quat2Like {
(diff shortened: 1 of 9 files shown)


src/quat2.ts
```diff
@@ -253,7 +253,7 @@ export class Quat2 extends Float32Array {
    * @param out - the receiving dual quaternion
    * @returns `out`
    */
-  static identity(out: QuatLike): QuatLike {
+  static identity<T extends Quat2Like>(out: T): T {
```
aidenfoxx, line 256: I assume it was a mistake that this was `QuatLike`?

PR #496, opened 2026-01-02
Branchless array initialization; faster rounding; other minor enhancements

solarunes:
## Array Initialization

Currently, all `create()` methods check `glMatrix.ARRAY_TYPE` to determine if the Array needs to manually be initialized to zero or not.

This can be avoided by creating a "zero-initializer class" that is a subclass of `glMatrix.ARRAY_TYPE`, if and only if it doesn't inherit a typed array.

### Implementation

I added a subclass generator for `Array`-like objects that adds zero initialization to their constructor. This subclass is available as a public field in the `glMatrix` namespace (`ARRAY_ZERO_INIT_TYPE`). 

Invoking `glMatrix.setMatrixArrayType(type)` now does the following:

- set `glMatrix.ARRAY_TYPE` to `type`
- if `type` is `Array`-like but not a typed array;
  - create a subclass `ArrayInitZero` of `type` that initializes it's elements to zero
  - set `ARRAY_ZERO_INIT_TYPE` to `ArrayInitZero`
- otherwise, set `ARRAY_ZERO_INIT_TYPE` to `type`

### Benchmarks

A full benchmark and comparison is available [here](https://jsperf.app/layuxu/5).

## Symmetric Rounding

The current ```glMatrix.round()``` implementation has branching- and modulo operations. None of this is required for symmetric rounding.

### Implementation

I replaced the old ```glMatrix.round()``` with a much faster, single line implementation, using only methods of the built-in ```Math``` namespace.

### Benchmarks

A full benchmark and comparison is available [here](https://jsperf.app/popoba/3).

## Other Enhancements

I shortened a few methods and introduced modern built-in functions where applicable. These should hopefully provide a minor performance boost.


psnet:
this AI code is awful.
one PR - one change

PR #498, opened 2026-06-27
Fix equals() returning true for vectors/values containing Infinity

### LINKED ISSUE #443: vector equals fails if one vector is all Infinity
```javascript
equals([1, 2, 3], [Infinity, Infinity, Infinity]) // should be false but is true
```

I think it's because of the way it multiplies `EPSILON` by the maximum value. This probably needs a special case.

Thanks!

chatman-media:
Closes #443.

`vec3.equals([1, 2, 3], [Infinity, Infinity, Infinity])` returns `true` (and matching infinities like `equals([Inf,Inf,Inf], [Inf,Inf,Inf])` return `false`).

### Cause
The per-component relative tolerance `EPSILON * Math.max(1, |a|, |b|)` evaluates to `Infinity` when an operand is infinite, so `Math.abs(a - b) <= Infinity` is satisfied for differing infinities and for finite-vs-infinite pairs. Conversely, two matching infinities give `Math.abs(Inf - Inf) === NaN`, and `NaN <= x` is `false`, so identical infinite vectors compare as *unequal*.

### Fix
For each component: short-circuit on exact equality (`a === b`, which also covers matching infinities) and clamp the relative scale with `Math.min(Number.MAX_VALUE, …)` so an infinite operand can no longer make the tolerance infinite. Results for all finite inputs are unchanged (for finite values the clamp is a no-op and the existing tests still pass).

Applied to `glMatrix.equals` (common.js) and `vec2`/`vec3`/`vec4` `equals` — the subject of the issue. The same inline pattern exists in `mat2`/`mat2d`/`mat3`/`mat4`/`quat`/`quat2`; happy to extend this PR to those too if you'd prefer them in one change.

### Tests
Added Infinity cases to the `equals` specs (common + vec2/vec3/vec4): finite-vs-infinite → `false`, matching infinities → `true`. Verified they fail on `master` and pass with the fix. Full suite: 1308 passing.

### DIFF
--- a/spec/gl-matrix/common-spec.js
+++ b/spec/gl-matrix/common-spec.js
@@ -14,19 +14,27 @@ describe("common", function(){
   });
 
   describe("equals", function() {
-    let r0, r1, r2, r3, r4;
+    let r0, r1, r2, r3, r4, r5, r6, r7, r8;
     beforeEach(function() {
       r0 = glMatrix.equals(1.0, 0.0);
       r1 = glMatrix.equals(1.0, 1.0);
       r2 = glMatrix.equals(1.0+glMatrix.EPSILON/2, 1.0);
       r3 = glMatrix.equals(1.0011, 1.0, 0.001);
       r4 = glMatrix.equals(100.5, 100.7, 0.2);
+      r5 = glMatrix.equals(1.0, Infinity);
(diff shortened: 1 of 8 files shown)


toji (approved):
Looks good, thank you!

PR #499, opened 2026-07-02
Clamp getAngle acos argument to avoid NaN

spokodev:
`quat.getAngle(a, b)` returns `NaN` for a large fraction of unit-quaternion inputs, including the angular distance of a quaternion with itself.

```js
const q = quat.setAxisAngle(quat.create(), [0, 0, 1], 0.7);
quat.getAngle(q, q); // NaN, expected ~0
```

The function computes `acos(2 * dot * dot - 1)`. For unit quaternions the dot product is ~1, and floating-point rounding can push it slightly above 1, so `2 * dot * dot - 1` exceeds 1 and `Math.acos` of an argument greater than 1 is `NaN`. In a sweep of unit quaternions this happens for about half of the `getAngle(q, q)` cases and for pairs a nanoradian apart (adjacent animation frames), and the `NaN` then makes downstream "are these orientations close" comparisons silently false.

## Fix

Clamp the `acos` argument to `[-1, 1]`, exactly as `vec3.angle` already does for the same hazard.

## Testing

Added a case to the existing `getAngle` "from itself" spec using a quaternion built from an axis and angle, which trips the rounding. It returns `NaN` on the current code and `0` with the fix. The existing `getAngle` specs still pass.

### DIFF
--- a/spec/gl-matrix/quat-spec.js
+++ b/spec/gl-matrix/quat-spec.js
@@ -516,6 +516,11 @@ describe("quat", function() {
           it("should be zero", function() {
               expect(quat.getAngle(quatA, quatA)).toBeEqualish(0);
           });
+
+          it("should be zero for a quaternion from an axis and angle", function() {
+              let q = quat.setAxisAngle(quat.create(), [0, 0, 1], 0.7);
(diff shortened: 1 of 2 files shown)


toji (approved):
LGTM, thanks for the fix!

PR #500, opened 2026-07-08
Fix vec3.slerp returning NaN for parallel inputs

spokodev:
`vec3.slerp` divided by `sin(angle)` with no guard, so when `a` and `b` point in the same direction the angle is zero, `sin(angle)` is zero and every output component became `0/0 = NaN`. For example `vec3.slerp(out, [0, 0, 1], [0, 0, 1], 0.5)` returned `[NaN, NaN, NaN]` instead of `[0, 0, 1]`.

The fix falls back to linear interpolation when the vectors are nearly parallel, which is the correct limit as the angle approaches zero. Adds a test for equal and parallel inputs.


toji:
Thank you!

PR #501, opened 2026-07-08
Fix mat4.getRotation and decompose under non-uniform scale

spokodev:
`mat4.getRotation` normalized the matrix by dividing each element by the scale of the wrong axis. Every column of a column-major mat4 must be divided by that column's scale, but the code paired `mat[1]`/`mat[2]` (still column 0) with `is2`/`is3` and `mat[4]`/`mat[8]` (columns 1 and 2) with `is1`. With a non-uniform scale this produced a non-unit, wrong quaternion; with uniform or no scale `is1 == is2 == is3` so the error was hidden.

For a 90 degree rotation about Z with scale `(2, 3, 4)`, `getRotation` returned `[0, 0, 0.766, 0.707]` with length 1.04 instead of the unit `[0, 0, 0.707, 0.707]`. This divides each column by its own scale so a rotation survives a `fromRotationTranslationScale` round trip. `decompose` duplicates the same code and is fixed together. Adds a test for each.

### DIFF
--- a/spec/gl-matrix/mat4-spec.js
+++ b/spec/gl-matrix/mat4-spec.js
@@ -634,6 +634,28 @@ function buildMat4Tests() {
                     expect(outangle).toBeEqualish(ang);
                 });
             });
+
+            describe("from a translation, rotation and non-uniform scale matrix", function() {
+                it("should keep the same rotation as when created", function() {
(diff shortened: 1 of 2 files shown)


toji:
And thank you again. 😄

